PR #48, opened 2021-11-25
add dummy property to test flux config

taomoh:
### JIRA link ###

https://tools.hmcts.net/jira/browse/ROC-9708

### Change description ###

Add dummy property to test flux config

```
[ ] Yes
[X] No
```

### DIFF
--- a/charts/civil-citizen-ui/values.preview.template.yaml
+++ b/charts/civil-citizen-ui/values.preview.template.yaml
@@ -3,4 +3,4 @@ nodejs:
   image: ${IMAGE_NAME}
   ingressHost: ${SERVICE_FQDN}
   environment:
-    - TEST_PROPERTY: test-value
+    TEST_PROPERTY: test-value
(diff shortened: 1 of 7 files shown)


charts/civil-citizen-ui/values.preview.template.yaml
```diff
@@ -1,4 +1,6 @@
-nodejs:
-  # Don't modify below here
-  image: ${IMAGE_NAME}
-  ingressHost: ${SERVICE_FQDN}
+    nodejs:
```
timja, line 1: indenting is broken in this file, remove the initial indent

config/custom-environment-variables.json
```diff
@@ -0,0 +1,17 @@
+{
```
timja, line 1: most other apps use yaml as its easier to read especially for config files.

taomoh: Can you provide an example of an app which uses yaml please? When I switched the format, the application broke locally. My understanding is that node.js expects environment variables to be defined in json files under config/ folder

vinothon84 (approved):
PR looks good and apprved now.

PR #58, opened 2021-12-06
Fix yarn audit vulnerabilities

Gulam-Hussain:
### JIRA link ### 
https://tools.hmcts.net/jira/browse/CIV-298

### Change description ###
I ran `yarn audit` and it displayed 20ish moderate vulnerabilities and 3 high vulnerabilities.
None of the packages that needed to be updated where inside the `package.json` and simply installing the specified version with the fix and then running `yarn install` did not solve the issue.
After some googling I came across [yarn resolutions](https://classic.yarnpkg.com/lang/en/docs/selective-version-resolutions/) as a solution which is what I have done for this PR.

```
[ ] Yes
[x] No
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -99,5 +99,15 @@
     "webpack-cli": "^4.9.1",
     "webpack-dev-middleware": "^5.2.2",
     "webpack-node-externals": "^3.0.0"
+  },
+  "resolutions": {
+    "**/chokidar/**/glob-parent": "^5.1.2",
+    "**/sass-lint/**/merge": "^2.1.1",
+    "**/uglifyjs-webpack-plugin/**/serialize-javascript": "^3.1.0",
+    "**/**/ansi-regex": "^5.0.1",
+    "**/jest/**/node-notifier": "^8.0.1",
+    "**/sass-lint/**/minimist": "^1.2.3",
+    "**/**/json-schema": "^0.4.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -99,5 +99,15 @@
     "webpack-cli": "^4.9.1",
     "webpack-dev-middleware": "^5.2.2",
     "webpack-node-externals": "^3.0.0"
+  },
+  "resolutions": {
+    "**/chokidar/**/glob-parent": "^5.1.2",
```
Gulam-Hussain, line 104: Where the package with the vulnerability was a sub-package of exactly one package, I used a path specific to that parent package, in this example, the package with the vulnerability was `glob-parent` and specifically under `chokidar` so the path used is as displayed

package.json
```diff
@@ -99,5 +99,15 @@
...
+    "**/sass-lint/**/merge": "^2.1.1",
+    "**/uglifyjs-webpack-plugin/**/serialize-javascript": "^3.1.0",
+    "**/**/ansi-regex": "^5.0.1",
```
Gulam-Hussain, line 107: For this example the package `ansi-regex` was causing a vulnerability in more than one specific package so I used a generic path instead.

johnoneill46 (approved):

faranhmcts (approved):

PR #94, opened 2022-01-13
Feature/civ 694 adjust dashboard UI to call civil api

taomoh:
### JIRA link ###

https://tools.hmcts.net/jira/browse/CIV-694

### Change description ###

1st vertical slice of "Claims made against you" dashboard

```
[ ] Yes
[X] No
```


src/main/app.ts
```diff
@@ -45,9 +45,7 @@ app.use((req, res, next) => {
 const testProperty = `${config.get<string>('testProperty')}`;
 logger.info(`Test Property value is: ${testProperty}`);
 
-glob.sync(__dirname + '/routes/**/*.+(ts|js)')
-  .map(filename => require(filename))
-  .forEach(route => route.default(app));
+app.use(dashboardRouter);
```
taomoh, line 48: The way cmc-citizen-frontend bulk-imports routes didn't work in the new code base. So I've used app.use(...routers). It would be good to make this automatically import all routes even nested ones under ./routes/**/*, without having to add them to app.use(...routes)

src/main/app/client/civilServiceClient.ts
```diff
@@ -0,0 +1,29 @@
...
+    const objects: Claim[] = response.data as Claim[];
+    const claims: Claim[] = [];
+    objects.map(item => {
+      const claim: Claim = new Claim();
+      claim.legacyCaseReference = item.legacyCaseReference;
+      claim.applicant1 = item.applicant1;
+      claim.totalClaimAmount = item.totalClaimAmount;
+      claim.respondent1ResponseDeadline = item.respondent1ResponseDeadline;
+      claims.push(claim);
+    });
```
taomoh, line 25: Is there a less verbose way to do this correctly? I've tried to use Object.assign() but the 2 functions defined within Claim class are then lost

src/main/common/models/claim.ts
```diff
@@ -0,0 +1,23 @@
...
+import format from '../../common/utils/moneyFormatter';
+
+export class Claim {
+  legacyCaseReference: string;
+  applicant1: Individual;
+  totalClaimAmount: number;
+  respondent1ResponseDeadline: Date;
+
+  formattedResponseDeadline(): string {
+    return this.respondent1ResponseDeadline ? dayjs(this.respondent1ResponseDeadline).format('D MMMM YYYY') : '';
```
taomoh, line 11: Is there a way to internationalise the date formatting? (e.g. display month in Welsh)

commitBlob: Should be able to do it via i18 library. Have you tried this [https://www.i18next.com/translation-function/formatting#datetime](https://www.i18next.com/translation-function/formatting#datetime)

src/main/app.ts
commitBlob: Unfortunately, I don't know any automated way to import routes, I would have to research that.
What we are doing is we have `routes.ts` file which contains anonymous function where we pass whole Application and add all routes and rules for each path. 
https://github.com/hmcts/pip-frontend/blob/master/src/main/routes/routes.ts

src/main/routes/features/dashboard/index.ts
```diff
@@ -0,0 +1,39 @@
+import * as express from 'express';
+import {CivilServiceClient} from '../../../app/client/civilServiceClient';
+import {Claim} from '../../../common/models/claim';
+
+const civilServiceApiBaseUrl = 'http://localhost:8765';
```
commitBlob, line 5: I would advise to create something like config file where you will set base url paths depending on the environment and import it anywhere you use it. That way you will only need to change the url in one place.
https://github.com/hmcts/pip-frontend/blob/master/src/main/resources/requests/utils/axiosConfig.ts

taomoh: This is a first iteration where the UI component is pointing to a wiremock instance locally to emulate an API endpoint which is still being developed. Once the endpoint available, we will change this to lookup the URL from an application property or an environment variable

src/main/app/client/civilServiceClient.ts
taomoh: For some reason, the spread operator doesn't map the 2 functions defined in the model class (`claim.ts`). I've used `Object.assign(new Claim(), _claim);` and that seems to map functions as well

src/main/app.ts
taomoh: I've took inspiration from you pip way and created a `routes.ts` file as well. Can you please review?

src/main/common/models/claim.ts
taomoh: I've tried this but I can't get it to work. I would like to show what I've attempted to see if you can something obvious that I am doing wrong

taomoh: @commitBlob as discussed earlier, we will attempt to resolve the date i18n problem at a later iteration, as it's not essential at this stage

commitBlob (approved):

vinothon84 (approved):

PR #109, opened 2022-01-27
Refactor/switch config format to yaml

taomoh:
### JIRA link ###

N/A

### Change description ###

Switched config file format from json to yaml. Also took this opportunity to add property for civil-service API base URL and 
to change port number from default template 3100 to 3001

```
[ ] Yes
[X] No
```


config/default.yaml
```diff
@@ -0,0 +1,18 @@
+security:
+  referrerPolicy: 'origin'
+port: 3001
+timeout: 30000
+services:
+  civilService:
+    url: 'http://localhost:8765'
+    microservice: 'civil_service'
+  idam:
+    authorizationURL: 'https://idam-web-public.aat.platform.hmcts.net/login'
```
mounikahmcts, line 10: To be consistent  we should point to localhost idam urls here as well. In custom environment variables we can have them pointed to respective variables via variables.

README.md
```diff
@@ -152,7 +150,7 @@ Make sure you have those values set correctly for your application.
 
 ### Healthcheck
 
-The application exposes a health endpoint (https://localhost:3100/health), created with the use of
+The application exposes a health endpoint (https://localhost:3001/health), created with the use of
```
LeonardoPalmeiro, line 153: why we have to change the port?

mounikahmcts (approved):

config/custom-environment-variables.yaml
```diff
@@ -0,0 +1,11 @@
+services:
```
LeonardoPalmeiro, line 1: 1) There is no   "useCSRFProtection": "true" 
2) are we mixing refactor with IDAM task?

config/default.yaml
```diff
@@ -0,0 +1,18 @@
+security:
```
LeonardoPalmeiro, line 1: there is no   "useCSRFProtection": true,

LeonardoPalmeiro (changes requested):

README.md
taomoh: 3100 is the default port provided by the expressjs template

config/custom-environment-variables.yaml
taomoh: There is no mix with IDAM task. The properties will be used by @osheaf01 and @laarmada once merged into master

LeonardoPalmeiro (approved):
perfect

PR #140, opened 2022-02-07
Civ 637 citizen sign in to c UI money claims portal

laarmada:
### JIRA link ###

https://tools.hmcts.net/jira/browse/CIV-637

### Change description ###
Allow sign-in to the created account using email address and password.

```
[x] Yes
[ ] No
```


charts/civil-citizen-ui/values.aat.template.yaml
```diff
@@ -2,3 +2,9 @@ nodejs:
   # Don't modify below here
   image: ${IMAGE_NAME}
   ingressHost: ${SERVICE_FQDN}
+
+idam-pr:
+  enabled: true
+  redirect_uris:
+    divorce:
```
taomoh, line 9: this isn't divorce

src/main/modules/oidc/index.ts
```diff
@@ -0,0 +1,52 @@
...
+    });
+
+    app.get(CALLBACK_URL, async (req: AppRequest, res: Response) => {
+      if (typeof req.query.code === 'string') {
+        req.session.user = await getUserDetails(redirectUri, req.query.code);
+        req.session.save(() => {
+          if (req.session.user?.roles?.includes('civil-citizen')) {
+            return res.redirect(DASHBOARD);
+          }
+          return res.render(UNAUTHORISED);
```
taomoh, line 28: render() takes a view name (unauthorised) and not a route (**/**unauthorised)

src/main/routes/urls.ts
```diff
@@ -0,0 +1,10 @@
+export type Urls = `/${string}`;
+
+export const CALLBACK_URL: Urls= '/oauth2/callback';
+export const SIGN_IN_URL: Urls = '/login';
+export const SIGN_OUT_URL: Urls = '/logout';
+export const DASHBOARD: Urls = '/dashboard';
+export const GET_CASES: Urls = '/cases';
```
taomoh, line 7: GET is HTTP method. Better to call this: CASES_URL

src/main/routes/urls.ts
```diff
@@ -0,0 +1,10 @@
+export type Urls = `/${string}`;
+
+export const CALLBACK_URL: Urls= '/oauth2/callback';
+export const SIGN_IN_URL: Urls = '/login';
+export const SIGN_OUT_URL: Urls = '/logout';
+export const DASHBOARD: Urls = '/dashboard';
```
taomoh, line 6: rename to DASHBOARD_URL to keep consistent with the other constant names

src/main/routes/urls.ts
```diff
@@ -0,0 +1,10 @@
...
+export const GET_CASES: Urls = '/cases';
+export const ROOT: Urls = '/';
+export const UNAUTHORISED: Urls = '/unauthorised';
```
taomoh, line 9: Append _URL to keep consistent with the other constant names

src/main/modules/oidc/index.ts
```diff
@@ -0,0 +1,52 @@
...
+
+    app.get(SIGN_IN_URL, (req: Request, res) => {
+      res.redirect(loginUrl + '?client_id=' + clientId + '&response_type=code&redirect_uri=' + encodeURI(redirectUri));
+    });
+
+    app.get(CALLBACK_URL, async (req: AppRequest, res: Response) => {
+      if (typeof req.query.code === 'string') {
+        req.session.user = await getUserDetails(redirectUri, req.query.code);
+        req.session.save(() => {
+          if (req.session.user?.roles?.includes('civil-citizen')) {
```
laarmada, line 25: we should include 'civil-citizen' into the config file as "default.yalm" file?

taomoh (changes requested):

src/main/routes/urls.ts
laarmada: done

src/main/routes/urls.ts
laarmada: done

src/main/routes/urls.ts
laarmada: done

src/main/modules/oidc/index.ts
laarmada: done

charts/civil-citizen-ui/values.aat.template.yaml
laarmada: done

src/main/app.ts
```diff
@@ -2,23 +2,42 @@ import * as bodyParser from 'body-parser';
...
 import { HealthCheck } from './modules/health';
+import { OidcMiddleware } from './modules/oidc';
 import routes from './routes/routes';
 
 const { Logger } = require('@hmcts/nodejs-logging');
 const { setupDev } = require('./development');
 
 const env = process.env.NODE_ENV || 'development';
 const developmentMode = env === 'development';
+export const cookieMaxAge = 21 * (60 * 1000); // 21 minutes
```
mounikahmcts, line 23: How did we come up with 21 min as maxCookieAge?

src/main/modules/oidc/index.ts
```diff
@@ -0,0 +1,52 @@
...
+            return res.redirect(DASHBOARD_URL);
+          }
+          return res.render('unauthorised');
```
mounikahmcts, line 28: Is it not "return res.redirect(UNAUTHORISED);"

src/test/unit/routes/home.test.ts
```diff
@@ -1,12 +1,35 @@
...
+      .then((res) => {
+        expect(res.status).toBe(302);
+      });
+}
+
 
 // TODO: replace this sample test with proper route tests for your application
 describe('Home page', () => {
+
+  const citizenRoleToken = 'eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJhY2Nlc3NUb2tlbiI6InRva2VuIiwic3ViIjoianNAdGVzdC5jb20iLCJnaXZlbl9uYW1lIjoiSm9obiIsImZhbWlseV9uYW1lIjoiU21pdGgiLCJ1aWQiOiIxMjMiLCJyb2xlcyI6WyJjaXZpbC1jaXRpemVuIl19.Ra3lo2bgl_mmiK8tHMVpBNf6mOTXDXUturb4Wy9ZbJc';
```
mounikahmcts, line 21: Is it possible to store the token in some test constants file so that we can reuse in all test files.

src/main/app.ts
taomoh: This was copied from nfdiv-frontend: https://github.com/hmcts/nfdiv-frontend/blob/18971ad9616085dd722ad66a0719e475b6dc819f/src/main/modules/session/index.ts#L12
Is there a specific requirement for Civil Citizen UI?

src/main/modules/oidc/index.ts
taomoh: I had res.redirect before but got a "too many redirect" exception. so switch to render

src/main/routes/urls.ts
taomoh: Not done for ROOT and UNAUTHORISED

src/main/modules/oidc/index.ts
taomoh: Good idea. Maybe somethings like:
services:
   idam:
      ....
      citizenRole: 'civil-citizen'

laarmada: exactly

src/main/app.ts
mounikahmcts: No I am not sure. Just checking if we are following same like other teams. I will mention it to steve when i demo things

LeonardoPalmeiro (approved):

config/custom-environment-variables.yaml
```diff
@@ -6,6 +6,7 @@ services:
     tokenURL: IDAM_API_URL
     clientID: IDAM_CLIENT_ID
     clientSecret: IDAM_SECRET
```
mounikahmcts, line 8: Are we setting idam secret any where?

config/default.yaml
```diff
@@ -5,13 +5,15 @@ port: 3001
...
+    url: 'http://localhost:4000'
     microservice: 'civil_service'
   idam:
     authorizationURL: 'http://localhost:3501/login'
     tokenURL: 'http://localhost:5000/o/token'
     clientID: 'civil_citizen_ui'
-    clientSecret: 'TBC'
+    clientSecret: 'citizen-ui-secret'
+    callbackURL: 'https://localhost:3001/oauth2/callback'
+    citizenRole: 'civil-citizen'
```
mounikahmcts, line 16: Where did we get this role?

laarmada: I have create this role, to allow civil-citizen-ui connect to idam. Else let me know which one is necessary

config/custom-environment-variables.yaml
laarmada: we should add to:
 values.preview.template.yaml file?
 values.yaml?

also IDAM_CLIENT_ID we haven´t ... please let me know which is the purpose of each yalm configuration file...

charts/civil-citizen-ui/values.yaml
```diff
@@ -2,10 +2,20 @@ nodejs:
...
 #      secrets:
 #        - AppInsightsInstrumentationKey
   environment:
-    TEST_PROPERTY: test-value-from-values-yaml
+    IDAM_WEB_URL: https://idam-web-public.{{ .Values.global.environment }}.platform.hmcts.net/login
+    IDAM_API_URL: https://idam-api.{{ .Values.global.environment }}.platform.hmcts.net/o/token
+    IDAM_CLIENT_ID: civil_citizen_ui
+    OAUTH_CLIENT_REDIRECT: https://civil-citizen-ui.{{ .Values.global.environment }}.platform.hmcts.net/oauth2/callback
+  autoscaling:
+    enabled: true
```
mounikahmcts, line 16: I think it should be disabled for preview and other environments. I see more pods are created in preview which will end up using all resources. We are already having resource issue on preview.
Screenshot 2022-02-09 at 09 21 15

config/custom-environment-variables.yaml
mounikahmcts: Yeah i was expecting this secret update somewhere in config. It is updated now. can resolve this comment.

config/default.yaml
mounikahmcts: I think this is not needed. All citizens will get citizen role by default when they create account. IF we create a new role then we have to assing that role to the user somehow when they are registering

config/default.yaml
```diff
@@ -5,13 +5,16 @@ port: 3001
 timeout: 30000
 services:
   civilService:
-    url: 'http://localhost:8765'
+    url: 'http://localhost:4000'
```
mounikahmcts, line 8: We are not overriding this value in custom env variables yaml so preview & other env fails.

laarmada: ok, fixed

charts/civil-citizen-ui/values.yaml
laarmada: I will set to false the autoscaling

mounikahmcts (approved):

PR #175, opened 2022-02-28
Civ 640 dashboard view cases

osheaf01:
### JIRA link ###
https://tools.hmcts.net/jira/browse/CIV-640

### Change description ###
Citizen Dashboard - citizen can view changes defended

```
[ ] Yes
[x] No
```


config/default.yaml
```diff
@@ -5,7 +5,7 @@ port: 3001
 timeout: 30000
 services:
   civilService:
-    url: 'http://localhost:8765'
```
laarmada, line 8: you shouldn´t change this value.

src/main/app/client/civilServiceClient.ts
```diff
@@ -1,37 +1,40 @@
...
   constructor(baseURL: string) {
-
     this.client = Axios.create({
       baseURL,
-
     });
   }
 
-  async retrieveByDefendantId(): Promise<Claim[]> {
-    const response: AxiosResponse<object> = await this.client.get('/cases');
```
laarmada, line 17: should we add an exception control here to avoid broken then webpage when the communication fails?

src/main/app/client/civilServiceClient.ts
```diff
@@ -1,37 +1,40 @@
...
-    const objects: Claim[] = response.data as Claim[];
-    const claims: Claim[] = [];
-    objects.forEach((_claim) => {
-      const claim: Claim = Object.assign(new Claim(), _claim);
-      claims.push(claim);
+  retrieveByDefendantId(req: AppRequest): Promise<Claim[]> {
+    return this.client.post(CASES_URL,
```
laarmada, line 19: I don´t like return large piece of code....

laarmada (changes requested):
ok

src/main/app/client/civilServiceClient.ts
```diff
@@ -1,37 +1,41 @@
...
+      return claims;
+    }).catch(error => {
+      console.log(error.message);
     });
-
-    return claims;
   }
 
   async retrieveClaimDetails(claimId: string): Promise<Claim> {
     const response: AxiosResponse<object> = await this.client.get(`/cases/${claimId}`);
```
laarmada, line 35: should we add an exception control here to avoid broken then webpage when the communication fails?

laarmada (changes requested):

src/main/app/client/civilServiceClient.ts
TatsyMaslova: I think this story is not about individual case. so the individual case retrieval is still broken. this story is to retrieve all cases. I think for the individual case we need a different story - as there are quite a bit of changes that are required and adding unit tests

config/default.yaml
TatsyMaslova: this is url for the civil service. if the civil service runs on port 4000 then the port previously specified is incorrect

laarmada: but in my last PR Mounika ask me that, so maybe it´ll fail on Jenkins

src/main/app/client/civilServiceClient.ts
TatsyMaslova: I agree. not very readable

config/default.yaml
TatsyMaslova: hmm perhaps we need to ask @mounikahmcts about this

src/main/app/client/civilServiceClient.ts
TatsyMaslova: also the method should be async because it returns a promise and also the client that is using the method needs to wait for the promise either to succeed or to fail

config/default.yaml
mounikahmcts: I have asked to update civilService url in preview/aat url of civil service in custom. environment variable like this 
https://github.com/hmcts/div-decree-nisi-frontend/blob/master/config/custom-environment-variables.yml#L67
Without which i cant test anything

src/main/app/client/civilServiceClient.ts
osheaf01: Have pushed changes to make this look nicer, please re-review.

src/main/app/client/civilServiceClient.ts
```diff
@@ -30,8 +41,6 @@ export class CivilServiceClient {
     if (!response.data) {
       throw new AssertionError({message: 'Claim details not available.'});
     }
```
LeonardoPalmeiro, line 43: we don't need    this if  if (!response.data) {...
because we're using .catch from post. 

My suggestion is:

    await this.client.post(CASES_URL,{ match_all: {} }, config)
      .then(response => {
        claims = response.data.cases.map((claim: CivilClaimResponse) => Object.assign(new Claim(), claim.case_data));
      }).catch(error => {
        throw new AssertionError({message: 'Claim details not available.'});
      });

src/main/app/client/civilServiceUrls.ts
```diff
@@ -0,0 +1 @@
+export const CASES_URL = '/cases/';
```
LeonardoPalmeiro, line 1: in this case we've a file to store all routerscan be found
in  src/main/routes/urls.ts

src/main/common/models/claim.ts
```diff
@@ -8,10 +8,8 @@ export class Claim {
   totalClaimAmount: number;
   respondent1ResponseDeadline: Date;
   detailsOfClaim: string;
-  respondent1: Respondent;
-  individualTitle: string;
-  individualLastName: string;
-  individualFirstName: string;
+  respondent1?: Respondent;
+
```
LeonardoPalmeiro, line 12: in the review version we had:
  individualTitle: string;
  individualLastName: string;
  individualFirstName: string;

do we remove it ?

src/main/routes/features/dashboard/index.ts
```diff
@@ -25,16 +25,13 @@ function renderPage(res: express.Response, claimsAsClaimant: Claim[], claimDraft
 const router = express.Router();
 
 router.get(DASHBOARD_URL, async function (req, res) {
+  const claimsAsClaimant : Claim[] = [];
```
LeonardoPalmeiro, line 28: do we need to send an empty object to the page?

LeonardoPalmeiro (changes requested):

src/main/app/client/civilServiceUrls.ts
TatsyMaslova: this is an api to civil service so I don't think it belongs in the routes/url.ts which is for routes that are for citizen-ui (i.e navigation through pages in browser)

src/main/common/models/claim.ts
TatsyMaslova: yes because they are repeating for respondents and applicant. the the json returned from api these are only defined in applicant and respondent . These are not needed  and are confusing where they belong to

src/main/routes/features/dashboard/index.ts
TatsyMaslova: yes, in my opinion because then the page will still be built even if there are no claims or if civil service is down

src/main/app/client/civilServiceClient.ts
TatsyMaslova: I think we are not changing this call for now. this call is for another story. the one that would deal with individual cases. this call as is will not work at the moment.

src/main/common/models/claim.ts
LeonardoPalmeiro: I agree with you, I was just make sure that not was a mistake

laarmada (approved):

PR #205, opened 2022-03-09
Civ 1525 redis claim service

TatsyMaslova:

### JIRA link ###

https://tools.hmcts.net/jira/browse/CIV-1525

### Change description ###

```
[ ] Yes
[x ] No
```

### DIFF
--- a/src/main/modules/draft-store/draftStoreService.ts
+++ b/src/main/modules/draft-store/draftStoreService.ts
@@ -3,25 +3,37 @@ import {CivilClaimResponse} from '../../common/models/civilClaimResponse';
 import {Claim} from '../../common/models/claim';
 
 export class DraftStoreService {
-
-  public async getDraftClaimFromStore(claimId: string): Promise<Claim> {
+  /**
+   * Gets civil claim response object with claim from draft store
+   * @param claimId
+   * @returns claim from redis or undefined when no there is no data for claim id
+   */
+  public async getDraftClaimFromStore(claimId: string): Promise<CivilClaimResponse> {
     const storedClaim: CivilClaimResponse = await app.locals.draftStoreClient.get(claimId);
-    return storedClaim?.case_data;
-  }
-
-  public async saveDraftClaim(claimId: string, claim:Claim) {
-    let storedClaim = await app.locals.draftStoreClient.get(claimId);
-    if(!storedClaim){
-      storedClaim = this.createNewClaim(claimId);
-    }
-    storedClaim.case_data = claim;
-    const draftStoreClient = app.locals.draftStoreClient;
-    draftStoreClient.set(claimId, storedClaim);
-  }
-
-  private createNewClaim(claimId: string) {
-    const storedClaim = new CivilClaimResponse();
-    storedClaim.id = claimId;
     return storedClaim;
+  }
+
+  /**
+   * Saves claim in Draft store. If the claim does not exist
+   * it creates a new CivilClaimResponse object and passes the claim in parameter to it
+   * then saves the new data in redis.
+   * If claim exists then it updates the claim with the new claim passed in parameter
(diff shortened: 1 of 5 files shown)


LeonardoPalmeiro (approved):
perfect

osheaf01 (approved):
LGTM

src/main/modules/draft-store/draftStoreService.ts
```diff
@@ -0,0 +1,27 @@
+import {app} from '../../app';
+import {CivilClaimResponse} from '../../common/models/civilClaimResponse';
+import {Claim} from '../../common/models/claim';
+
+export class DraftStoreService {
+
+  public async getDraftClaimFromStore(claimId: string): Promise<Claim> {
+    const storedClaim: CivilClaimResponse = await app.locals.draftStoreClient.get(claimId);
+    return storedClaim?.case_data;
```
sampankumar, line 9: Does this return null when draftStore doesn't have the case?

src/main/modules/draft-store/draftStoreService.ts
```diff
@@ -0,0 +1,27 @@
...
+  }
+
+  public async saveDraftClaim(claimId: string, claim:Claim) {
+    let storedClaim = await app.locals.draftStoreClient.get(claimId);
+    if(!storedClaim){
+      storedClaim = this.createNewClaim(claimId);
```
sampankumar, line 15: Should we not call civil-service here first to check if the claim exist in ccd or not? If it does then we would want to use that.

src/main/modules/draft-store/draftStoreService.ts
```diff
@@ -0,0 +1,27 @@
...
+import {Claim} from '../../common/models/claim';
+
+export class DraftStoreService {
+
+  public async getDraftClaimFromStore(claimId: string): Promise<Claim> {
+    const storedClaim: CivilClaimResponse = await app.locals.draftStoreClient.get(claimId);
+    return storedClaim?.case_data;
+  }
+
+  public async saveDraftClaim(claimId: string, claim:Claim) {
```
laarmada, line 12: should create other one for statementOfMeans data Model? or we will refactor when the data structure will be defined completely?

laarmada (changes requested):
Only some doubts

src/main/modules/draft-store/draftStoreService.ts
LeonardoPalmeiro, line 9: maybe the best option would be return all the storedClaim object instead .case_data.
when the application is up there is a clam mock from redisData.json

LeonardoPalmeiro (changes requested):

src/main/modules/draft-store/draftStoreService.ts
TatsyMaslova: this is not about our changes for statementOfMeans this is about claim. But I believe statementOfMean is part of the claim if you look at ocmc thats what they do

src/main/modules/draft-store/draftStoreService.ts
TatsyMaslova: I think this service is only about draftStore and should not have any knowledge of ccd. In my opinion we should have another orchestrating service that when required would retrieve the claim from ccd and store it in redis with the help of this service. At the moment in our code the call to ccd to get individual claim is not working. so we should address that in a different story. I think this particular task and pr is just to do with the draft store (redis)

src/main/modules/draft-store/draftStoreService.ts
TatsyMaslova: hmm yes, i think you are right. in case we need id for example and then other possible fields that would be added

LeonardoPalmeiro (approved):

PR #206, opened 2022-03-09
Civ 1525 redis claim service

TatsyMaslova:

### Change description ###

```
[ ] Yes
[ ] No
```

### DIFF
--- a/src/main/modules/draft-store/draftStoreService.ts
+++ b/src/main/modules/draft-store/draftStoreService.ts
@@ -3,25 +3,37 @@ import {CivilClaimResponse} from '../../common/models/civilClaimResponse';
 import {Claim} from '../../common/models/claim';
 
 export class DraftStoreService {
-
-  public async getDraftClaimFromStore(claimId: string): Promise<Claim> {
+  /**
+   * Gets civil claim response object with claim from draft store
+   * @param claimId
+   * @returns claim from redis or undefined when no there is no data for claim id
+   */
+  public async getDraftClaimFromStore(claimId: string): Promise<CivilClaimResponse> {
     const storedClaim: CivilClaimResponse = await app.locals.draftStoreClient.get(claimId);
-    return storedClaim?.case_data;
-  }
-
-  public async saveDraftClaim(claimId: string, claim:Claim) {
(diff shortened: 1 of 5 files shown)


src/main/modules/draft-store/draftStoreService.ts
```diff
@@ -22,18 +24,26 @@ export class DraftStoreService {
...
+    draftStoreClient.set(claimId, JSON.stringify(storedClaimResponse));
   }
 
   private createNewCivilClaimResponse(claimId: string) {
     const storedClaimResponse = new CivilClaimResponse();
     storedClaimResponse.id = claimId;
     return storedClaimResponse;
   }
+
+  private convertRedisData(data:any): CivilClaimResponse{
```
LeonardoPalmeiro, line 42: that we have in Redis is String of representation of json, might be better change data:any to data: string,

LeonardoPalmeiro: maybe the best name for this method could be convertRedisDataToCivilClaimResponse or something like this. what do you think about it?

src/main/modules/draft-store/draftStoreService.ts
```diff
@@ -22,18 +24,26 @@ export class DraftStoreService {
...
+    let jsonData = undefined;
+    if(data){
+      jsonData = JSON.parse(data);
+    }
+    return jsonData? Object.assign(jsonData, new CivilClaimResponse()): undefined;
```
LeonardoPalmeiro, line 47: I believe that jsonData will be always null in this case. 

we need all methods on from claims class like formattedResponseDeadline formattedTotalClaimAmount, in this case my suggestion is 
    return Object.assign(new CivilClaimResponse(), jsonData); 
in this case if we don't have the jsonData will create a empty object of CivilClaimResponse.

LeonardoPalmeiro (changes requested):

src/main/modules/draft-store/draftStoreService.ts
TatsyMaslova: done 👍

laarmada (approved):
Amazing!

LeonardoPalmeiro (approved):
great!!

PR #207, opened 2022-03-11
Civ 1263 list bancks savings accounts

TatsyMaslova:

### Change description ###

```
[ ] Yes
[ ] No
```


src/main/assets/js/append-row.js
```diff
@@ -0,0 +1,60 @@
+document.addEventListener('DOMContentLoaded', function () {
+  const appendRowButton = document.getElementsByClassName('append-row');
+  if (appendRowButton && appendRowButton.length > 0) {
+    appendRowButton[0].addEventListener('click', (event) => {
+      event.preventDefault();
+      cloneRow();
+    });
+  }
+
+  function cloneRow() {
```
sampankumar, line 10: Will it be beneficial to make this method parameterised, where class name can be passed may be?

src/main/assets/js/append-row.js
```diff
@@ -0,0 +1,60 @@
...
+    if (document.getElementsByClassName('multiple-row') && document.getElementsByClassName('multiple-row').length > 0) {
+      const lastRow = getLastRow();
+      let newRow = lastRow.cloneNode(true);
+      lastRow.parentNode.appendChild(newRow);
+      updateNewRow();
+    }
+  }
+
+  function getLastRow(){
```
sampankumar, line 19: Will it be beneficial to make this method parameterised, where class name can be passed may be?

src/main/assets/js/append-row.js
TatsyMaslova: this happens on dom /page load similar to Rosario's approach. so it works so that if we want to add multiple rows then we just use specific class to the button and place our row in a div with a specific class.

src/main/assets/js/append-row.js
TatsyMaslova: this happens on dom /page load similar to Rosario's approach. so it works so that if we want to add multiple rows then we just use specific class to the button and place our row in a div with a specific class.

TatsyMaslova: so it is reusable - we reuse classes rather then the method. as it is called on load and assigns the functionality to the button with the class

src/test/unit/routes/features/response/statementOfMeans/bankAccounts/bankAccountsController.test.ts
```diff
@@ -0,0 +1,168 @@
...
+      };
+      await request(app).post(getBaseUrlWithIdParam(CLAIM_ID) + CITIZEN_BANK_ACCOUNT_URL)
+        .send(data)
+        .expect((res) => {
+          expect(res.status).toBe(200);
+          expect(res.text).toContain(SELECT_AN_OPTION);
+        });
+    });
+    test('should return error when type of account is not specified', async ()=>{
+      const data = {
```
laarmada, line 59: should we have a const data in a external file?....

src/main/views/features/response/statementOfMeans/citizenBankAndSavings/citizen-bank-accounts.njk
```diff
@@ -0,0 +1,60 @@
...
+      {% endif %}
+      <div class="govuk-grid-row">
+        <div class="govuk-grid-column-two-thirds">
+          <h1 class="govuk-heading-l">{{ t('List your bank and savings accounts') }}</h1>
+        </div>
+      </div>
+      <div class="govuk-grid-row">
+        <div class="govuk-grid-column-two-thirds">
+          <form name="accounts" method="post">
+              <p class="govuk-body">
```
laarmada, line 25: could we think in a generic component? to create a generic macro to reuse it?

src/main/routes/features/response/statementOfMeans/bankAccounts/bankAccountsController.ts
```diff
@@ -0,0 +1,50 @@
...
+});
+
+async function renderErrorsIfExist(form: BankAccounts, res: express.Response, claimId:string) {
+  await validateForm(form);
+  await validateFormArray(form.accounts);
+  if (form.hasErrors()) {
+    renderView(form, new BankAccountTypes(), res);
+  } else {
+    await bankAccountService.saveBankAccounts(claimId, form);
+    res.redirect(getBaseUrlWithIdParam(claimId) + CITIZEN_SEVERELY_DISABLED_URL);
```
laarmada, line 46: I will change the method including the concat url to avoid "+", for example: getBaseUrlWithIdParam(claimId,CITIZEN_SEVERELY_DISABLED_URL)

src/main/common/form/models/bankAndSavings/bankAccountType.ts
```diff
@@ -0,0 +1,14 @@
+import {BankAccountTypeValues} from './bankAccountTypeValues';
+
+export class BankAccountType {
+
+  readonly value: BankAccountTypeValues;
+  readonly text: string;
+  selected: boolean;
```
laarmada, line 7: in which case is used "selected" parameter?

laarmada (changes requested):
we need discuss about that.

LeonardoPalmeiro (changes requested):

src/main/views/features/response/statementOfMeans/citizenBankAndSavings/citizen-bank-accounts.njk
TatsyMaslova: do you mean like a generic macro for adding lots of rows? i don't think it is possible. however I do use a macro for account data which can be reusable

src/main/common/form/models/bankAndSavings/bankAccountType.ts
TatsyMaslova: when there is existing data and the data value is the same as the bank account types then the selected parameter is set to true so that we can see it selected in the drop down

src/main/routes/baseUrlConstants.ts
```diff
@@ -0,0 +1 @@
+export const BASE_CASE_RESPONSE_URL = '/case/:id/response';
```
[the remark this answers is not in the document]
TatsyMaslova: done :) good spot

src/main/common/utils/urlFormatter.ts
```diff
@@ -0,0 +1,3 @@
+export function getBaseUrlWithIdParam(id: string): string{
+  return `/case/${id}/response`;
```
LeonardoPalmeiro, line 2: should use information from constant, but probably will came with another problem which is replace the **:id**.
I did something in the https://github.com/hmcts/civil-citizen-ui/blob/676363d7ebe6f14ca11c760e35bed28d625f87d3/src/main/common/utils/urlPatchReplace.ts#L1.

LeonardoPalmeiro (changes requested):

src/main/modules/statementOfMeans/bankAccounts/bankAccountService.ts
```diff
@@ -0,0 +1,37 @@
+import {BankAccounts} from '../../../common/form/models/bankAndSavings/bankAccounts';
+import {BankAccount} from '../../../common/form/models/bankAndSavings/bankAccount';
+import {getCaseDataFromStore, saveDraftClaim} from '../../../modules/draft-store/draftStoreService';
+import {Claim} from '../../../common/models/claim';
+import {StatementOfMeans} from '../../../common/models/statementOfMeans';
+import {convertFormToCitizenBankAccount, convertCitizenBankAccountsToForm} from './bankAccountConverter';
+
+export class BankAccountService {
+  public async getBankAccounts(claimId: string) {
```
laarmada, line 9: we should add the return type?

laarmada (changes requested):
other thing

src/test/unit/routes/features/response/statementOfMeans/bankAccounts/bankAccountsController.test.ts
TatsyMaslova: it is debatable. because in this case data for every test is different. and also one would want to see what data is there. if i put it in a different file it would sort of hide the actual data when you look at the test. So I think in this case it should be visible in the test rather then be in the external file

src/main/assets/js/append-row.js
sampankumar: Can this then also be used to add lets say another dropdown? or is it just for the button, because we are using the button css class?

TatsyMaslova: it has nothing to do with drop downs - its only to do with button that would "clone" all elements that are in specific div. and it assigns this functionality after dom elements are loaded

TatsyMaslova: so if we have a button and it has a specific class then it will have the functionality to clone elements in the div with also specific class. it works similar to Rosario's lookup code where the functionality is assigned to a button with a specific class

src/main/routes/features/response/statementOfMeans/bankAccounts/bankAccountsController.ts
TatsyMaslova: done :)

src/test/unit/routes/features/response/statementOfMeans/bankAccounts/bankAccountsController.test.ts
laarmada: Maybe we can create a generic data but we don´t have enough time.... thanks @TatsyMaslova

laarmada (approved):
great job.

PR #222, opened 2022-03-15
Feature/civ 1266 where do you live screen

taomoh:
### JIRA link ###

https://tools.hmcts.net/jira/browse/CIV-1266

### Change description ###

Add `Where do you live` screen

```
[ ] Yes
[X] No
```


src/test/a11y/ignored-urls.ts
```diff
@@ -26,4 +26,8 @@ export const IGNORED_URLS =
     urls.CITIZEN_PARTENER_PENSION_URL,
     urls.CITIZEN_DEPENDANTS_URL,
     urls.CITIZEN_PARTNER_DISABILITY_URL,
+    urls.CITIZEN_PARTNER_AGE_URL,
+    urls.CITIZEN_PARTENER_PENSION_URL,
```
GorilaNaranja, line 30: Duplicated:
CITIZEN_PARTENER_PENSION_URL
CITIZEN_DEPENDANTS_URL
CITIZEN_PARTNER_DISABILITY_URL

src/main/routes/features/response/statementOfMeans/residenceRoute.ts
```diff
@@ -0,0 +1,69 @@
...
+          form: form,
+        });
+      } else {
+        saveResidence(req.params.id, residence).then(() => {
+          res.redirect(CITIZEN_PARTNER_URL);
+        });
+      }
+    });
+
+const saveResidence = async (claimId: string, residence: Residence): Promise<void> => {
```
laarmada, line 52: we are using a service class to have this kind of Business logic

src/main/routes/features/response/statementOfMeans/residenceRoute.ts
```diff
@@ -0,0 +1,69 @@
...
+import {CivilClaimResponse} from '../../../../common/models/civilClaimResponse';
+import {StatementOfMeans} from '../../../../common/models/statementOfMeans';
+
+const {Logger} = require('@hmcts/nodejs-logging');
+
+const logger = Logger.getLogger('residenceRoute');
+const residenceViewPath = 'features/response/statement-of-means/residence';
+
+const residenceRoute = express.Router();
+residenceRoute
```
laarmada, line 18: I think we should choose the correct way to do that... with .get....then().post  or using await/async.... but I think we should use the same structure all developers....

src/main/routes/features/response/statementOfMeans/residenceRoute.ts
```diff
@@ -0,0 +1,69 @@
+import express from 'express';
```
laarmada, line 1: we have *Controller at the end of that files....

src/main/common/models/statementOfMeans.ts
```diff
@@ -2,9 +2,12 @@ import {Disability} from '../form/models/statementOfMeans/disability';
 import {SevereDisability} from '../form/models/statementOfMeans/severeDisability';
 import { Partner } from '../form/models/statementOfMeans/partner';
 import {CitizenBankAccount} from './citizenBankAccount';
+import {Residence} from '../form/models/statementOfMeans/residence';
+
 export class StatementOfMeans {
   disability?: Disability;
   severeDisability?: SevereDisability;
   partnerAge?: Partner;
```
laarmada, line 10: it should be a  PartnerAge instead of Partner

laarmada (changes requested):
we can discuss about that comments

src/main/routes/features/response/statementOfMeans/residenceRoute.ts
taomoh: This will be revisited one Sampan has finalised the draft response model

src/main/routes/features/response/statementOfMeans/residenceRoute.ts
taomoh: The root folder is called routes. I think the terms controllers and routs are being used interchangeably

src/main/common/models/statementOfMeans.ts
taomoh: this isn't my code

src/main/routes/features/response/statementOfMeans/residenceRoute.ts
taomoh: I am happy to revisit this once we agree on the approach

src/main/routes/features/response/statementOfMeans/residenceRoute.ts
```diff
@@ -0,0 +1,69 @@
...
+    (req: express.Request, res: express.Response) => {
+      getDraftClaimFromStore(req.params.id).then((draftResponse: CivilClaimResponse) => {
+        assert(draftResponse);
+
+        res.render(residenceViewPath, {
+          form: new GenericForm(draftResponse.case_data.statementOfMeans.residence),
+        });
+      });
+    })
+  .post(
```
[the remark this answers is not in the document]
taomoh: I am happy to revisit this once we agree on the approach

src/test/a11y/ignored-urls.ts
taomoh: Well spotted. removed duplicates

src/main/routes/features/response/statementOfMeans/residenceRoute.ts
laarmada: so we should change all other class names to avoid inconsistency ....

src/main/common/models/statementOfMeans.ts
laarmada: ok...

src/main/routes/features/response/statementOfMeans/residenceRoute.ts
taomoh: My understanding is that using services to store the draft response was still not a final decision. The team may come up with a different approach.

laarmada (approved):
I´m no happy to avoid different project structure and names.... but I´ll approve.

src/main/routes/features/response/statementOfMeans/residenceRoute.ts
taomoh: the object exported by controller files is built using `express.Router()` . I will change to controller for consistency

GorilaNaranja (approved):
Great

src/main/routes/features/response/statementOfMeans/residenceRoute.ts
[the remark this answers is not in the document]
taomoh: it doesn't seem to be possible to call then() or catch() on router.post()

laarmada (commented):
thanks Tao for the changes!

LeonardoPalmeiro (approved):

PR #284, opened 2022-03-30
Civ 1360 children live with you disabled

osheaf01:
### JIRA link ###

https://tools.hmcts.net/jira/browse/CIV-1360

### Change description ###

Asking if children are disabled screen - but only when certain preceding conditions are met.

```
[ ] Yes
[x ] No
```


src/main/modules/statementOfMeans/dependants/childrenDisabilityService.ts
```diff
@@ -12,7 +12,8 @@ export const isCheckChildrenDisabled = (claim: Claim): boolean => {
   try {
     const statementOfMeans = claim.statementOfMeans;
     if (statementOfMeans?.dependants?.numberOfChildren && totalNumberOfChildren(statementOfMeans.dependants.numberOfChildren) > 0) {
-      if ((statementOfMeans?.disability?.option == YesNo.NO) || (statementOfMeans?.disability?.option == YesNo.YES && statementOfMeans?.severeDisability?.option == YesNo.NO && statementOfMeans?.partnerDisability?.option != YesNo.YES)){
+      if ((statementOfMeans?.disability?.option == YesNo.NO) || (statementOfMeans?.disability?.option == YesNo.YES &&
```
TatsyMaslova, line 15: this is a bit hard to read. perhaps you could put the business logic into the model rather then in the service. for example put the knowlege whether the disability option is set to yes in the disability model for example disability.isOptionYesSelected(). The same with severeDisability and partnerDisablitity then in your service you can just return without if eg. return statementOfMeans?.disability?.isOptionYesSelected || (statementOfMeans?.disability?.isOptionYesSelected && statementOfMeans?.severedisability?.isOptionYesSelected && s(statementOfMeans?.partnerDisability?.isOptionYesSelected )

src/main/common/form/models/statementOfMeans/dependants/numberOfChildren.ts
```diff
@@ -34,7 +34,21 @@ export class NumberOfChildren {
   }
 }
 
+export function totalNumberOfChildren(numberOfChildren: NumberOfChildren) {
```
TatsyMaslova, line 37: not sure it is clean to put an export function inside a class. could be a method inside a class and have that business logic inside the model

TatsyMaslova: in addition this needs a unit test :)

src/main/modules/statementOfMeans/dependants/betweenSixteenAndNineteenService.ts
```diff
@@ -26,19 +26,21 @@ export const getForm = async (claimId: string): Promise<BetweenSixteenAndNinetee
...
 
-export const saveFormToDraftStore = async (claimId: string, form: BetweenSixteenAndNineteenDependants) => {
+export const saveFormToDraftStore = async (claimId: string, form: BetweenSixteenAndNineteenDependants): Promise<Claim> => {
   try {
     const claim = await getClaim(claimId);
     const statementOfMeans = claim.statementOfMeans ? claim.statementOfMeans : new StatementOfMeans();
     statementOfMeans.numberOfChildrenLivingWithYou = form.value;
     claim.statementOfMeans = statementOfMeans;
     await saveDraftClaim(claimId, claim);
+    return claim;
```
TatsyMaslova, line 36: Why should it return claim?

src/main/modules/statementOfMeans/dependants/dependantsService.ts
```diff
@@ -35,6 +36,7 @@ class DependantsService {
         civilClaimResponse.case_data.statementOfMeans = statementOfMeans;
       }
       await saveDraftClaim(claimId, civilClaimResponse.case_data);
+      return civilClaimResponse.case_data;
```
TatsyMaslova, line 39: again i don't think it should return anything when it is saving data.

src/main/modules/statementOfMeans/dependants/betweenSixteenAndNineteenService.ts
osheaf01: Controller needs Claim to decide which redirect to use.
I can look at moving the logic about asking whether to ask if children are disabled into Claim itself.

src/main/modules/statementOfMeans/dependants/childrenDisabilityService.ts
osheaf01: Unit tests added. Will refactor to make it slightly easier to read. 
It's not a terribly simple algorithm.

src/main/modules/statementOfMeans/dependants/childrenDisabilityService.ts
```diff
@@ -0,0 +1,89 @@
...
+
+export function setChildrenDisabilityServiceLogger(winstonLogger: winston.LoggerInstance) {
+  logger = winstonLogger;
+}
+
+export const hasDisabledChildren = (claim: Claim): boolean => {
+  try {
+    const statementOfMeans = claim.statementOfMeans;
+    const numberOfChildren : NumberOfChildren = new NumberOfChildren(statementOfMeans?.dependants?.numberOfChildren?.under11, statementOfMeans?.dependants?.numberOfChildren?.between11and15, statementOfMeans?.dependants?.numberOfChildren?.between16and19);
+    return (statementOfMeans?.dependants?.numberOfChildren && numberOfChildren.totalNumberOfChildren() > 0 &&
```
TatsyMaslova, line 23: you only need to check whether you numberOfChildren is not undefined. so i think it would be `numberOfChildren?.totalNumberOfChildren()>0 && ...`

osheaf01: I'll check and rerun

src/main/modules/statementOfMeans/dependants/childrenDisabilityService.ts
```diff
@@ -0,0 +1,89 @@
...
+  } catch (error) {
+    logger.error(error);
+    throw error;
+  }
+};
+
+export const validateChildrenDisability = (childrenDisability: ChildrenDisability): GenericForm<ChildrenDisability> => {
+  try {
+    const form: GenericForm<ChildrenDisability> = new GenericForm(childrenDisability);
+    form.errors = validator.validateSync(form.model);
```
TatsyMaslova, line 80: in my opinion validation can be inside the generic form. because it contains model. and also i don't think it should be validateSync. so we could have a const validator outside the generic form and then a method validate but that is just my opionion. For example inside GenericModel:
` public async validate() {
    this.errors = await validator.validate(this.model as unknown as object);
  }`

osheaf01: Moved to genericForm.

TatsyMaslova (approved):

src/main/common/form/models/statementOfMeans/dependants/numberOfChildren.ts
```diff
@@ -32,9 +32,25 @@ export class NumberOfChildren {
...
+      total += this?.between16and19;
+    }
+    return total;
+  }
 }
 
+
+
 function toNumberOrUndefined(value: string): number {
   const numberValue: number = parseFloat(value);
```
LeonardoPalmeiro, line 54: maybe in this part we could use parseInt
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt
what do you think ?

LeonardoPalmeiro (changes requested):

LeonardoPalmeiro (changes requested):
there are some spots untested.
image

src/main/common/form/models/statementOfMeans/dependants/numberOfChildren.ts
osheaf01: I thought so, too, I think it's definitely an Integer. Taoufik wrote this so reluctant to change it too much.

LeonardoPalmeiro (approved):

PR #320, opened 2022-04-13
Feature/civ 1366 court orders

taomoh:
### JIRA link ###

https://tools.hmcts.net/jira/browse/CIV-1366

### Change description ###

Add "Court orders" screen

```
[ ] Yes
[X] No
```


src/main/routes/features/response/statementOfMeans/courtOrders/courtOrdersController.ts
```diff
@@ -0,0 +1,52 @@
...
+      }
+    })
+  .post(
+    CITIZEN_COURT_ORDERS_URL,
+    async (req: express.Request, res: express.Response) => {
+      const courtOrders = courtOrdersService.buildCourtOrders(req.body);
+      courtOrdersService.removeEmptyCourtOrders(courtOrders);
+      const form: GenericForm<CourtOrders> = courtOrdersService.validateCourtOrders(courtOrders);
+
+      if (form.hasErrors() || form.hasNestedErrors()) {
```
TatsyMaslova, line 37: has errors will return true when there are only child errors 😉 so no need to call form.hasNestedErrors()

src/main/routes/features/response/statementOfMeans/courtOrders/courtOrdersController.ts
```diff
@@ -0,0 +1,52 @@
...
+        logger.error(`${error.stack || error}`);
+        res.status(500).send({errorMessage: error.message, errorStack: error.stack});
+      }
+    })
+  .post(
+    CITIZEN_COURT_ORDERS_URL,
+    async (req: express.Request, res: express.Response) => {
+      const courtOrders = courtOrdersService.buildCourtOrders(req.body);
+      courtOrdersService.removeEmptyCourtOrders(courtOrders);
+      const form: GenericForm<CourtOrders> = courtOrdersService.validateCourtOrders(courtOrders);
```
TatsyMaslova, line 35: we have form.validate() method. so you can do await form.validate() instead of putting validation in the service. i think service layer should not really know about validation /view/request/response etc

src/main/services/features/response/statementOfMeans/courtOrders/courtOrdersService.ts
```diff
@@ -0,0 +1,59 @@
...
+
+  public buildCourtOrders(value: unknown): CourtOrders {
+    return CourtOrders.fromObject(value as Record<string, object>);
+  }
+
+  public removeEmptyCourtOrders(courtOrders: CourtOrders): void {
+    courtOrders.rows = courtOrders.rows.filter(item => !item.isEmpty());
+  }
+
+  public validateCourtOrders(courtOrders: CourtOrders): GenericForm<CourtOrders> {
```
TatsyMaslova, line 51: as mentioned above there is already a validate method in generic form so you don't need that code here.

src/main/views/features/response/statementOfMeans/courtOrders/court-orders.njk
```diff
@@ -0,0 +1,100 @@
...
+
+              <div class="govuk-form-group">
+                {{ govukButton({
+                  text: t("Save and continue")
+                }) }}
+              </div>
+
+            </form>
+          </div>
+
```
TatsyMaslova, line 90: sorry for being picky - don't think you need that extra space

src/main/routes/features/response/statementOfMeans/courtOrders/courtOrdersController.ts
taomoh: Good point I will change this

src/main/services/features/response/statementOfMeans/courtOrders/courtOrdersService.ts
taomoh: see my response above. Plus I need to build the form from the model as I am using the generic form. This happens inside validateCourtOrders() function

src/main/views/features/response/statementOfMeans/courtOrders/court-orders.njk
taomoh: You're right. I've deleted it

src/main/routes/features/response/statementOfMeans/courtOrders/courtOrdersController.ts
TatsyMaslova: in addition with having validation in the form we don't have to repeat the code again and again (where we set the form.errors to errors returned by the validator etc.) so we avoid code repetition (as well as hiding the knowledge of what happens with errors etc inside the form)

src/main/services/features/response/statementOfMeans/courtOrders/courtOrdersService.ts
TatsyMaslova: i think i prefer not to repeat code over like this. i prefer that generic form knows that it has to set its errors after validation. it also knows about the model. so hiding the knowledge of what happens during validation inside the form in my opinion is nicer. so then you wouldn't need this method. all you would need in the controller is : const form = new GenericlForm(model);
await form.validate();
if(form.hasErrors()){
....
}

if we put the logic outside the form then we have to repeat this code with form.errors = validator.validate(model) in different places. especially if we are adding this in service layer for each service. A lot of code repetition

taomoh: I've tried this and it makes sense. The only reservation I had was about having the validator inside the model. In Java world, usually the validator is external to the bean: https://www.baeldung.com/javax-validation#2-validate-the-bean
But I guess if we are using the generic form approach, it's not a conflict.
I've pushed the changes

src/main/routes/features/response/statementOfMeans/courtOrders/courtOrdersController.ts
taomoh: Resolving this as covered by another comment

src/main/services/features/response/statementOfMeans/courtOrders/courtOrdersService.ts
```diff
@@ -0,0 +1,50 @@
...
+import {CourtOrders} from '../../../../../common/form/models/statementOfMeans/courtOrders/courtOrders';
+
+const {Logger} = require('@hmcts/nodejs-logging');
+const logger = Logger.getLogger('courtOrdersService');
+
+class CourtOrdersService {
+
+  public async getCourtOrders(claimId: string): Promise<CourtOrders> {
+    try {
+      const civilClaimResponse = await getDraftClaimFromStore(claimId);
```
laarmada, line 12: if you use getCaseDataFromStore instead of getDraftClaimFromStore you save a lot of code in this class...

laarmada (commented):
great Job Tao!!

TatsyMaslova (approved):
Great work!!

src/main/services/features/response/statementOfMeans/courtOrders/courtOrdersService.ts
taomoh: You're right @laarmada . I've changed to getCaseDataFromStore

src/main/services/features/response/statementOfMeans/courtOrders/courtOrdersService.ts
```diff
@@ -22,15 +22,16 @@ class CourtOrdersService {
 
   public async saveCourtOrders(claimId: string, courtOrders: CourtOrders) {
     try {
-      const civilClaimResponse = await getDraftClaimFromStore(claimId);
-      if (civilClaimResponse?.case_data?.statementOfMeans) {
-        civilClaimResponse.case_data.statementOfMeans.courtOrders = courtOrders;
+      const claim = await getDraftClaimFromStore(claimId);
+
+      if (claim?.statementOfMeans) {
```
laarmada, line 27: We can change this part: 
if (!claim.statementOfMeans) { 
 claim.statementOfMeans = new StatementOfMeans();
}
claim.statementOfMeans.courtOrders = courtOrders;

taomoh: I like it. I've changed it. Thanks!

laarmada (approved):

commitBlob (approved):

PR #321, opened 2022-04-13
New: Carer page

GorilaNaranja:
### JIRA link ###
https://tools.hmcts.net/jira/browse/CIV-1699

### Change description ###
Carer’s Allowance or Carer’s Credit

```
[ ] Yes
[x] No
```


src/main/routes/features/response/statementOfMeans/carerController.ts
```diff
@@ -0,0 +1,46 @@
...
+  } catch (error) {
+    logger.error(error);
+    res.status(500).send({ error: error.message });
+  }
+});
+
+carerController.post(CITIZEN_CARER_URL,
+  async (req, res) => {
+    const carer: Carer = new Carer(req.body.option);
+    const errors: ValidationError[] = validator.validateSync(carer);
```
TatsyMaslova, line 31: I think we have a validator that sets the errors for you in the form that we could use :) so you don't need to do it in the controller

src/main/modules/statementOfMeans/carerService.ts
```diff
@@ -0,0 +1,38 @@
...
+import {getCaseDataFromStore, saveDraftClaim} from '../draft-store/draftStoreService';
+import {StatementOfMeans} from '../../common/models/statementOfMeans';
+
+const {Logger} = require('@hmcts/nodejs-logging');
+const logger = Logger.getLogger('carerService');
+
+export const getCarer = async (claimId: string) => {
+  try {
+    const case_data = await getCaseDataFromStore(claimId);
+    if (case_data.statementOfMeans?.carer) {
```
osheaf01, line 11: Was a discussion on my most recent PR - whether to use case_data or camelCase.
Perhaps call it "claim"?

src/main/views/features/response/statementOfMeans/carer.njk
```diff
@@ -0,0 +1,37 @@
...
+      {% set errors = form.getErrors() %}
+      {% if form.hasErrors() %}
+        {{ govukErrorSummary({
+          "titleText": t("There was a problem"),
+          "errorList": errors
+        }) }}
+      {% endif %}
+      <div class="govuk-grid-row">
+        <div class="govuk-grid-column-two-thirds">
+          <form method="post">
```
osheaf01, line 23: Add "novalidate" to the post?

src/main/modules/statementOfMeans/carerService.ts
commitBlob: Agree with this, for the consistency I would argue to rename the function as well to `getClaimDataFromStore` although this would affect a bit of files

laarmada (approved):
great job!

src/main/modules/statementOfMeans/carerService.ts
GorilaNaranja: @commitBlob I´m agree with you, I would use `getClaimFromStore`. I think if we want to rename that function we should create another PR for that.

laperla11 (approved):

PR #337, opened 2022-04-20
feat(local): mock sign in

commitBlob:
### JIRA link ###
n/a

### Change description ###
Added environment variable to 'flip' between sign ins as I noticed that if we don't have idam service running we only comment out block of code. With the introduction of the environment variable we don't need to modify any of the code, it is only required to add environment variable (`MOCK_SIGN_IN=true`)

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/modules/oidc/index.ts
+++ b/src/main/modules/oidc/index.ts
@@ -44,13 +44,17 @@ export class OidcMiddleware {
     });
 
     app.use((req: AppRequest, res: Response, next: NextFunction) => {
-      if (req.session.user) {
-        if (req.session?.user?.roles?.includes(citizenRole)) {
-          return next();
+      if (process.env.MOCK_SIGN_IN === 'true') {
+        return next();
+      } else {
+        if (req.session.user) {
+          if (req.session?.user?.roles?.includes(citizenRole)) {
+            return next();
+          }
+          return res.redirect(DASHBOARD_URL);
         }
-        return res.redirect(DASHBOARD_URL);
+        res.redirect(SIGN_IN_URL);
       }
-      res.redirect(SIGN_IN_URL);
     });
   }
 }


src/main/modules/oidc/index.ts
```diff
@@ -44,13 +44,17 @@ export class OidcMiddleware {
     });
 
     app.use((req: AppRequest, res: Response, next: NextFunction) => {
-      if (req.session.user) {
-        if (req.session?.user?.roles?.includes(citizenRole)) {
-          return next();
+      if (process.env.MOCK_SIGN_IN === 'true') {
```
taomoh, line 47: I am wondering if this could be maliciously exploited to short-circuit authentication, for example by setting the environment variable on a test or production environment. Probably not, as Azure infrastructure is supposed to prevent this type of unauthorised actions

laarmada: As I have ask in slack, from my point of view, thats changes could be use in local environment but it hasn´t sense in master... thats only my opinion.

commitBlob: The environment variables are set when the process starts (when application boots up via Jenkins deployment). We are using server side rendering therefore code comes compiled to the browser and we don't have access to environment variables there as they are already injected. Since we don't define `MOCK_SIGN_IN ` environment variable on the pipeline it will not be set, therefore the if block I added will never be triggered.
There are examples of these environment variables set on other projects
https://github.com/search?q=org%3Ahmcts+process.env&type=code
https://github.com/hmcts/pip-frontend/blob/master/src/main/app.ts#L26
https://github.com/hmcts/pip-frontend/blob/ee0433bfab0bb2010a1fc5590b617760b5efa50d/src/main/routes/routes.ts#L17
Capture

commitBlob:
closing since there is a pushback from the team

PR #346, opened 2022-04-22
Civ 1815 task list model

GorilaNaranja:
### JIRA link ###
https://tools.hmcts.net/jira/browse/CIV-1815

### Change description ###
Response Task List model

### Comments ###

The Gov UK is working on the component Task List Page. We are importing their scss file (that might not be finished) into our main.scss just to have an example of how it will looks like.

[Task list pages Component](https://design-system.service.gov.uk/patterns/task-list-pages/)

This files are just examples and we will need to work on them in nexts issues:
- src/main/assets/scss/task-list.scss
- src/main/common/utils/taskListBuilder.ts
- src/main/common/utils/tasks/confirmYourDetails.ts
- src/main/common/utils/tasks/needMoreTime.ts
- src/main/modules/taskListService.ts
- src/main/routes/features/response/taskListController.ts
- src/main/views/features/response/task-list.njk

```
[x] Yes
[ ] No
```


src/main/modules/taskListService.ts
```diff
@@ -0,0 +1,43 @@
+import {TaskStatus} from "../common/form/models/TaskStatus";
```
taomoh, line 1: I know there are already tons of service classes under module folder. but ideally these should be under src/main/services

src/main/modules/taskListService.ts
```diff
@@ -0,0 +1,43 @@
...
+import {TaskList} from "../common/form/models/taskList";
+
+/**
+ * THIS FILE IS A CONCEPT
+ * 
+ * This code is only a concept of what we should do. 
+ * 
+ */
+
+let data: TaskList[] = [];
```
taomoh, line 11: The name `data` is too abstract in this context. This should be renamed to be more descriptive: `taskLists`, for example?

src/main/modules/taskListService.ts
```diff
@@ -0,0 +1,43 @@
...
+let completed = 0;
+let total = 0;
+
+const calculateTotalAndCompleted = (tasksLists: TaskList[]) => {
+  data = [];
+  completed = 0;
+  total = 0;
+  tasksLists.forEach(tasksList => {
```
taomoh, line 19: Can `completed` not be calculated in this loop as well? this would might save some precious milliseconds :)

src/main/modules/taskListService.ts
```diff
@@ -0,0 +1,43 @@
...
+  });
+
+  data.forEach(taskList => {
+    completed += countCompletedTasks(taskList);
+  });
+}
+
+const getTitle = (tasksLists: TaskList[]) => {
+  calculateTotalAndCompleted(tasksLists);
+  return completed < total ? 'Application incomplete' : 'Application complete';
```
taomoh, line 31: How would i18n work with this?

src/main/modules/taskListService.ts
```diff
@@ -0,0 +1,43 @@
...
+}
+
+const getDescription = (tasksLists: TaskList[]) => {
+  calculateTotalAndCompleted(tasksLists);
+  return `You have completed ${completed} of ${total} sections`;
```
taomoh, line 36: How would i18n work with this?
In the Java world I would use this: https://stackoverflow.com/a/2205654/2219361
Maybe this is too much for the scope of this PR but we have to address this at some point

src/main/routes/features/response/taskListController.ts
```diff
@@ -0,0 +1,61 @@
...
+    data.push(
+      taskListPrepareYourResponse,
+      taskListRespondeToClaim,
+      taskListTryToResolveClaim,
+      taskListYourHearingRequirements,
+    );
+    data = data.filter(item => item.tasks.length !== 0);
+    const title = getTitle(data);
+    const description = getDescription(data);
+
```
taomoh, line 49: Maybe this logic should go to the service layer?

taomoh (commented):
Great job @GorilaNaranja . I've added some comments nevertheless :)

src/main/common/utils/taskListBuilder.ts
```diff
@@ -0,0 +1,73 @@
+import {TaskStatus} from "../../common/form/models/TaskStatus";
```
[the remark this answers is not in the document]
GorilaNaranja: The taskListBuilder.ts is the common file that will build all the sets of TaskList.
The files under "common/utils/tasks", are the builders for each Task and we will have one for each different Task.

src/main/modules/taskListService.ts
GorilaNaranja: Yes but I didn't want to make more complex the structure of the project and I decided to put it on the folder where all the services are for now. We can change it at any moment.

src/main/modules/taskListService.ts
GorilaNaranja: This is not the scope of the issue and also I don't know to much about how multilanguage is working on this project. In other projects I worked on we did something like that (I worked with i18n in Vue.js but the fundamentals should be the same):

We have 2 yaml or json files:

en.yaml
```yaml
app_messages:
  app_complete: Application incomplete
  app_incomplete: Application complete
```

es.yaml
```yaml
app_messages:
  app_complete: Aplicación incompleta
  app_incomplete: Aplicación completa
```

i18n does the logic to detect the language and use en.yaml or es.yaml file. Pretty similar to what we have now on i18n/index.ts file:
` loadPath: __dirname + '/locales/{{lng}}/{{ns}}.json',`

And in in our views or ts files we can use this:
`t(app_messages.app_complete)`

src/main/modules/taskListService.ts
GorilaNaranja: In this case I found this on stackoverflow but I didn't test it:

You can do that like this (example):

`I18n.t('helpers.links.add', variable: "Your Variable Value")`

then, on your translation file (yml):

```yaml
helpers:
  links:
    add: "Add %{variable}"
```

[StackOverflow](https://stackoverflow.com/questions/45511052/i18n-attaching-variable-to-translation-string?msclkid=f3966b47c24111ecbe00e0f8533e7c14)

sonar-project.properties
```diff
@@ -5,5 +5,5 @@ sonar.sources=src/main
 sonar.tests=src/test/
 sonar.test.exclusions=test/coverage/**
 sonar.javascript.lcov.reportPaths=coverage/lcov.info
-sonar.exclusions=src/main/app.ts,src/main/development.ts,src/main/index.js,src/main/server.ts,src/main/assets/js/postcode-lookup.js,src/main/assets/js/calculate-amount.js
+sonar.exclusions=src/main/app.ts,src/main/development.ts,src/main/index.js,src/main/server.ts,src/main/assets/js/postcode-lookup.js,src/main/assets/js/calculate-amount.js,src/main/assets/scss/task-list.scss,src/main/common/utils/taskList/taskListBuilder.ts,src/main/common/utils/taskList/tasks/confirmYourDetails.ts,src/main/common/utils/taskList/tasks/needMoreTime.ts,src/main/modules/taskListService.ts,src/main/routes/features/response/taskListController.ts
```
taomoh, line 8: I assume that once different scenarios will start being developed later on, the files you added to sonar.exclusions should be removed as they should be test-covered. If that is the case, I suggest putting a TODO comment on top of this property so we don't forget about that

src/main/modules/taskListService.ts
taomoh: Probably one for the tech debt backlog

src/main/modules/taskListService.ts
taomoh: You're right we can work that out when we start focusing on i18n aspect

src/main/modules/taskListService.ts
taomoh: Something to pick with i18n work

LeonardoPalmeiro (approved):

taomoh (approved):
Great stuff @GorilaNaranja !

TatsyMaslova (approved):

RosDiaf (approved):

src/main/common/utils/taskList/taskListBuilder.ts
```diff
@@ -0,0 +1,73 @@
...
+  status: TaskStatus.COMPLETE,
+};
+
+const howMuchYouHavePaidTask = {
+  description: 'Tell us how much you’ve paid',
+  url: '/how-much-paid',
+  status: TaskStatus.INCOMPLETE,
+};
+
+const buildPrepareYourResponseSection = async (claim: Claim): Promise<TaskList> => {
```
laperla11, line 28: I know this is a concept but we don't have any `async` ops inside this function, maybe it is better to remove to prevent any confusion.

src/main/common/utils/taskList/taskListBuilder.ts
```diff
@@ -0,0 +1,73 @@
...
+  const needMoreTimeTask = getneedMoreTimeTask(claim);
+
+  tasks.push(confirmYourDetailsTask);
+  if (claim.respondent1ResponseDeadline > new Date()) {
+    tasks.push(needMoreTimeTask);
+  }
+  return { title: 'Prepare your response', tasks };
+};
+
+const buildRespondeToClaimSection = async (claim: Claim): Promise<TaskList> => {
```
laperla11, line 40: same here

src/main/common/utils/taskList/taskListBuilder.ts
```diff
@@ -0,0 +1,73 @@
...
+  const tasks: Task[] = [];
+
+  tasks.push(chooseAResponseTask);
+  if (!claim.paymentOption) {
+    tasks.push(howMuchYouHavePaidTask);
+  }
+  return { title: 'Responde to Claim', tasks };
+};
+
+const buildTryToResolveClaimSection = async (claim: Claim): Promise<TaskList> => {
```
laperla11, line 50: same here

src/main/common/utils/taskList/taskListBuilder.ts
```diff
@@ -0,0 +1,73 @@
...
+  const tasks: Task[] = [];
+
+  if (!claim.paymentOption) {
+    tasks.push(howMuchYouHavePaidTask);
+  }
+  return { title: 'Try to resolve the Claim', tasks };
+};
+
+const buildYourHearingRequirementsSection = async (claim: Claim): Promise<TaskList> => {
```
laperla11, line 59: same here

src/main/modules/taskListService.ts
```diff
@@ -0,0 +1,64 @@
...
+ * THIS FILE IS A CONCEPT
+ * 
+ * This code is only a concept of what we should do. 
+ * 
+ */
+
+let completed = 0;
+let total = 0;
+
+const getTaskLists = async (claim: Claim) => {
```
laperla11, line 21: if you decided to remove async from build functions so we should remove the async from here as well.

src/main/routes/features/response/taskListController.ts
```diff
@@ -0,0 +1,35 @@
...
+const { Logger } = require('@hmcts/nodejs-logging');
+const logger = Logger.getLogger('taskListController');
+const taskListViewPath = 'features/response/task-list';
+const taskListController = express.Router();
+
+taskListController.get(CLAIM_TASK_LIST_URL, async (req, res) => {
+  try {
+    const claim: Claim = await getDraftClaimFromStore(req.params.id);
+    const taskLists = await getTaskLists(claim);
```
laperla11, line 24: if you remove the previous `async` s we need to remove await from here.

laperla11 (approved):

PR #379, opened 2022-04-28
Fix/civ 2044 duplicate error message

taomoh:
### JIRA link ###

https://tools.hmcts.net/jira/browse/CIV-2044

### Change description ###

Fix bug of duplicate/unwanted error messages in "Who employs you" screen and refactor it to use `GenericForm`

```
[ ] Yes
[X] No
```


src/main/common/utils/xpathExtractor.ts
```diff
@@ -0,0 +1,16 @@
+//https://developer.mozilla.org/en-US/docs/Web/API/XPathResult
+const ORDERED_NODE_SNAPSHOT_TYPE = 7;
```
TatsyMaslova, line 2: would this be only used in test? if so perhaps we should have it in utils in test folder

taomoh: Yes. I've moved it across

TatsyMaslova (approved):

LeonardoPalmeiro (approved):

PR #380, opened 2022-04-29
CIV-1981 - Free Telephone Mediation

commitBlob:
### JIRA link ###
https://tools.hmcts.net/jira/browse/CIV-1981

### Change description ###
Implement free telephone mediation page, which redirects applicants depending on their type
Individual and Sole Trader are redirected to `confirm-your-telephone` page while Company and Organisation are redirected to `can-we-use-company`
Added macro to set page titles dynamically instead of static titles
Fixed minor typos in the taskListBuilder

```
[ ] Yes
[x] No
```


src/main/routes/features/response/mediation/freeTelephoneMediationController.ts
```diff
@@ -0,0 +1,21 @@
...
+import {getDraftClaimFromStore} from '../../../../modules/draft-store/draftStoreService';
+
+const freeTelephoneMediationController = express.Router();
+const citizenFreeTelephoneMediationViewPath = 'features/response/mediation/free-telephone-mediation';
+
+freeTelephoneMediationController.get(CITIZEN_FREE_TELEPHONE_MEDIATION_URL, async (req, res) => {
+  try {
+    const civilClaim = await getDraftClaimFromStore(req.params.id);
+    res.render(citizenFreeTelephoneMediationViewPath,
+      { isBusiness: (civilClaim.case_data.applicant1.type === 'ORGANISATION' || civilClaim.case_data.applicant1.type === 'COMPANY') },
```
laarmada, line 12: it will be fail if the case_data is empty you should check it before to access to that variables.
Also you can send to njk the civilClaim directly and check there....

src/main/routes/features/response/mediation/freeTelephoneMediationController.ts
```diff
@@ -0,0 +1,21 @@
...
+    );
+  }
+  catch (error) {
+    res.status(500).send({error: error.message});
+  }
+
+});
+
```
laarmada, line 20: you should add the navigation logic. (Post?)
We have several scenarios not cover for this controller. For example:

Scenario: Save and continue + Yes + Claimant type = Company or Organisation
Given citizen is viewing 'You chose not to try free mediation' screen
When Save and continue selected
And 'Yes' selected
And defendant type = Organisation' Or 'Company'
Then display 'Who should the mediation service call' screen IF contact name for company or organisation is Null 
Or display 'The right person for mediation service to call' screen IF contact name for company or organisation is Not Null

src/main/routes/features/response/mediation/freeTelephoneMediationController.ts
laarmada: I just to check it:
image

laarmada (changes requested):

src/main/views/macro/page-title.njk
```diff
@@ -0,0 +1,6 @@
+{% macro setPageTitle(title)%}
+  {% set dynamicTitle = t('Your money claims account') + ' - ' + t(title) if title else t('Your money claims account - Money Claims') %}
+  {% block pageTitle %}
+    {{ dynamicTitle }}
```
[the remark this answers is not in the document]
commitBlob: this is for the tab title, not the page title
pageTitle

src/main/routes/features/response/mediation/freeTelephoneMediationController.ts
commitBlob: But you should not be able to see this page at all if there is no applicant as this page is displayed when you click an item from the task list

laarmada: if for any reason the applicant hasn´t type... it will be fail, we have the applicant but without type....

src/main/routes/features/response/mediation/freeTelephoneMediationController.ts
commitBlob: There is no save and continue here? I am not updating any data on this page

laarmada (approved):

src/main/routes/urls.ts
```diff
@@ -1,5 +1,6 @@
 const BASE_CASE_RESPONSE_URL = '/case/:id/response';
 const STATEMENT_OF_MEANS_URL = `${BASE_CASE_RESPONSE_URL}/statement-of-means`;
+const MEDIATION_URL = '/case/:id/mediation';
```
GorilaNaranja, line 3: Check this file to avoid conflicts, we are using this approach in a couple of PRs [url file](https://github.com/hmcts/civil-citizen-ui/blob/c440adb2de0934d078a87d311e9db846d2c7e63a/src/main/routes/urls.ts)

laarmada: done :)

PR #394, opened 2022-05-05
New: confirm details for company and organisation

GorilaNaranja:
### JIRA link ###
https://tools.hmcts.net/jira/browse/CIV-2252

### Change description ###
Confirm Details - Name and Address - Company or Organisation

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/modules/draft-store/redisData.json
+++ b/src/main/modules/draft-store/redisData.json
@@ -31,6 +31,8 @@
       },
       "respondent1": {
         "type": "ORGANISATION",
+        "companyName": "Version 1",
+        "contactPerson": "Test contact person",
         "partyName": "Example respondent1 organisation",
         "primaryAddress": {
           "County": "Kent",
@@ -693,6 +695,8 @@
       },
       "respondent1": {
         "type": "COMPANY",
+        "companyName": "Version 1",
+        "contactPerson": "Test contact person",
         "partyName": "Example respondent1 organisation",
         "primaryAddress": {
           "County": "Kent",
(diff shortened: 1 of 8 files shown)


src/main/modules/draft-store/redisData.json
```diff
@@ -693,6 +695,8 @@
       },
       "respondent1": {
         "type": "COMPANY",
+        "companyName": "Version 1",
+        "contactPerson": "Test contact person",
```
TatsyMaslova, line 699: would that mean that when you go to the page where it asks to specify the name of the contact, the input field will be prepopulated with the contact person? shouldn't we have one that does not have this information? because i think the information is not coming from ccd for example when there is an existing claim just raised. the information comes from citizen ui

GorilaNaranja: Yes, it means that. This claim will prepopulate the "contactPerson" just as an example of what happens when we have that data already stored in Redis but I'm going to delete it. I already set "contactPerson" for the claim with id 1645882162449409 so we don't need both.

TatsyMaslova: cool 😃

TatsyMaslova (approved):

src/main/views/features/response/citizenDetails/citizen-details-company.njk
```diff
@@ -0,0 +1,138 @@
...
+            {% set noHtml %}
+            {% endset -%}
+            {% set correspondenceAddressHtml %}
+            <div class="govuk-form-group" id="postcode">
+              <p class="govuk-label govuk-!-margin-bottom-5">{{ t('All correspondence will be sent to this address. This is known as an "address for service".') }}</p>
+              <label class="govuk-label" for="postcode">
+                {{ t('Enter a UK postcode') }}
+              </label>
+              <div id="postcode-error" class="govuk-visually-hidden govuk-error-message">
+                We can’t find an address for that postcode. Try again or enter the address manually.
```
RosDiaf, line 66: Translation (t) missing here?

RosDiaf (approved):

laarmada (approved):

src/main/views/features/response/citizenDetails/citizen-details-company.njk
GorilaNaranja: Yes, I'm going to change it here and also in citizen-details.njk file

PR #482, opened 2022-05-31
add redis async method

LeonardoPalmeiro:

### Change description ###
fix when redis is getting the old information.

```
[ ] Yes
[x] No
```

### DIFF
--- a/infrastructure/main.tf
+++ b/infrastructure/main.tf
@@ -22,9 +22,24 @@ module "civil-citizen-ui-draft-store" {
   common_tags  = var.common_tags
 }
 
+module "citizen-ui-draft-store" {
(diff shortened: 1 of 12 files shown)


commitBlob (approved):

laarmada (approved):

taomoh (changes requested):
1. I don't think we need to handle the connect event explicitly as the call `new Redis()` will create the connection as explained here: https://github.com/luin/ioredis#basic-usage and here: https://github.com/luin/ioredis#connect-to-redis. IMO, this adds unnecessary overhead. Can you please explain what problem you're trying to solve by adding the connect event handler and how you think this is going to resolve it?

2. How come yarn.lock changed without package.json having been modified?

PR #493, opened 2022-06-09
Civ 2588 fees client

TatsyMaslova:

### JIRA link ###
https://tools.hmcts.net/jira/browse/CIV-2588

### Change description ###

```
[ ] Yes
[x ] No
```


src/main/app/client/civilServiceClient.ts
```diff
@@ -55,17 +57,13 @@ export class CivilServiceClient {
...
-      { claimAmountRange: '£5,000.01 to £10,000', fee: '£455' },
-    ];
-    return feesRanges;
+  async getFeeRanges(req: AppRequest): Promise<FeeRange[]> {
+    const config = this.getConfig(req);
+    try{
+      const response: AxiosResponse<object> = await this.client.get(CIVIL_SERVICE_FEES_URL, config);
+      return plainToInstance(FeeRange, response.data as object[]);
+    } catch (err: unknown) {
+      logger.error(err);
```
taomoh, line 66: What would the user see if there is an error caught while retrieving the fee ranges?
Should the error be propagated, and then caught and rendered in the generic error page?
Maybe we can just replicate current OCMC behaviour

src/main/app/client/civilServiceUrls.ts
```diff
@@ -1 +1,2 @@
 export const CIVIL_SERVICE_CASES_URL = '/cases/';
+export const CIVIL_SERVICE_FEES_URL = '/fees/';
```
taomoh, line 2: If you action my comment in the civil-service PR, this will have to change accordingly

src/main/common/models/feeRange.ts
```diff
@@ -1,5 +1,53 @@
+import 'reflect-metadata';
```
taomoh, line 1: I don't see this module being explicitly used. Is this an implicit dependency for class transformation?

src/main/common/models/feeRange.ts
```diff
@@ -1,5 +1,53 @@
...
+  readonly currentVersion: CurrentVersion;
+
+  constructor (minRange: number, maxRange: number, currentVersion: CurrentVersion) {
+    this.minRange = minRange;
+    this.maxRange = maxRange;
+    this.currentVersion = currentVersion;
+  }
+
+  formatFeeRangeToTableItem():TableItem[] {
+    return [{text: `£${this.minRange.toLocaleString()} to £${this.maxRange.toLocaleString()}`}, {text: `£${this.currentVersion.flatAmount.amount.toLocaleString()}`}];
```
taomoh, line 21: I assume this will have to support i18n. Maybe not in scope of current story?

src/main/common/models/feeRange.ts
```diff
@@ -1,5 +1,53 @@
...
+  @Type(() => CurrentVersion)
+  readonly currentVersion: CurrentVersion;
+
+  constructor (minRange: number, maxRange: number, currentVersion: CurrentVersion) {
+    this.minRange = minRange;
+    this.maxRange = maxRange;
+    this.currentVersion = currentVersion;
+  }
+
+  formatFeeRangeToTableItem():TableItem[] {
```
taomoh, line 20: I would personally have had the formatting logic in the view. that way you wouldn't need to reference the translation function in the model class

src/test/unit/app/client/civilServiceClient.test.ts
```diff
@@ -4,7 +4,7 @@ import {Claim} from '../../../../main/common/models/claim';
 import * as requestModels from '../../../../main/common/models/AppRequest';
 import {CivilClaimResponse} from '../../../../main/common/models/civilClaimResponse';
 import config from 'config';
-import {CIVIL_SERVICE_CASES_URL} from '../../../../main/app/client/civilServiceUrls';
+import {CIVIL_SERVICE_CASES_URL, CIVIL_SERVICE_FEES_URL} from '../../../../main/app/client/civilServiceUrls';
```
taomoh, line 7: multiple lines will help avoid future merge conflicts

taomoh (changes requested):

src/main/common/models/feeRange.ts
TatsyMaslova: I prefer my entity to be intelligent as per DDD so that it returns the string and knows the ins and outs of it. whereas the controller would not care what is there as long as it returns the type it wants to send to the view

taomoh: I guess it's a matter of personal preference. I still think this is a presentation concern which should be delegated to the view.
Also how would you manage translation in that case without injecting the translation logic into the entity itself? Unless you think translation can take place in the entity itself

TatsyMaslova: I think it can and we already have that in the code where we pass the language

taomoh: I haven't seen in an entity in our code yet. Could you point me to an example.
I know it can theoretically be done in the entity, I meant is it sensible to have translation at the entity level?

src/main/app/client/civilServiceClient.ts
```diff
@@ -3,8 +3,10 @@ import Axios, {AxiosInstance, AxiosResponse} from 'axios';
 import {AssertionError} from 'assert';
 import {AppRequest} from '../../common/models/AppRequest';
 import {CivilClaimResponse} from '../../common/models/civilClaimResponse';
-import {CIVIL_SERVICE_CASES_URL} from './civilServiceUrls';
+import {CIVIL_SERVICE_CASES_URL, CIVIL_SERVICE_FEES_URL} from './civilServiceUrls';
```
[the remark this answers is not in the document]
TatsyMaslova: done :)

src/main/common/models/feeRange.ts
TatsyMaslova: ah we have done it in a service. but i don't see problem doing it in a entity. this is how i actually realised that we can do it in the backend - by seeing an example in an entity from someone else. I don't think though formatting to string is something that it is defining how it looks in the view. it does not contain html it just returns a type and values in that type.

src/main/app/client/civilServiceClient.ts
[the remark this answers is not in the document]
taomoh: i don't see it. maybe you haven't pushed to remote?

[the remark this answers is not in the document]
TatsyMaslova: I can see: 
image

laarmada (approved):

src/main/common/models/feeRange.ts
taomoh: If you do translation in the view, you wouldn't even need to worry about passing the language to the translation function. You can just call it this way: `t('COMMON.FEE_RANGE_WORDING', {min: model.min, max: model.max})`

Assuming you have an entry in the translation JSON files like this: `"FEE_RANGE_WORDING": "{{min}} to {{max}}"`. See this for more details: https://www.i18next.com/translation-function/interpolation

If you do it in the model itself, you will have to:
- import the translation function in the model class
- add a language argument to the translation function call
- reference the translation key
	
```
import {t} from 'i18next';
formatFeeRangeToTableItem(lang: string):TableItem[] {
   t('COMMON.FEE_RANGE_WORDING', {min: model.min, max: model.max, lng: getLng(lang)});
}
```

Doing this at the model is controversial, because the model is supposed to be lightweight (think POJO in Java world), ideally with zero-interaction with other non model classes/modules. Doing it at service level is not as controversial because the service naturally does the heavy lifting and orchestration which could involve interaction with model and non-model classes/modules.

BTW, presentation is not necessarily synonymous to HTML. For example, you could present the same fee range in multiple ways:
- 10 to 20
- From 10 to 20
- Between 10 and 20
- 10 - 20
The Prepositions above and the hyphen are presentation artefacts, whereas the numbers 10 and 20 are model properties

src/main/app/client/civilServiceClient.ts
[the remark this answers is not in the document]
taomoh: Please have the curly brackets separate from the constants as, otherwise, conflicts could still occur if they're on the same line:
```
import {
    CIVIL_SERVICE_CASES_URL,
    CIVIL_SERVICE_FEES_URL,
} from ...
```

taomoh (approved):

PR #516, opened 2022-06-17
Civ 3042 fix translation with unexpected data

TatsyMaslova:

### JIRA link ###

https://tools.hmcts.net/jira/browse/CIV-3042

### Change description ###
some additional changes to cope with unexpected data that is returned by the fees api

```
[ ] Yes
[ x] No
```

### DIFF
--- a/src/main/routes/features/response/eligibility/sendYourResponseByEmailController.ts
+++ b/src/main/routes/features/response/eligibility/sendYourResponseByEmailController.ts
@@ -43,10 +43,23 @@ sendYourResponseByEmailController.get(SEND_RESPONSE_BY_EMAIL_URL, async (req, re
 
 const formatFeesRanges = (feesRanges: FeeRange[]): [TableItem[]] => {
   const tableFormatFeesRanges: [TableItem[]] = [[]];
-  feesRanges.forEach((feeRange: FeeRange) => {
-    tableFormatFeesRanges.push(feeRange.formatFeeRangeToTableItem());
+  let previousFeeRange: FeeRange = undefined;
+  feesRanges.sort((element1:FeeRange, element2:FeeRange ) => {
+    if(element1.maxRange < element2.maxRange){
+      return -1;
+    }
+    if(element1.maxRange > element2.maxRange){
+      return 1;
+    }
+    return 0;
+  }).forEach((feeRange: FeeRange) => {
+    if(!feeRange.equals(previousFeeRange)){
+      tableFormatFeesRanges.push(feeRange.formatFeeRangeToTableItem());
+    }
+    previousFeeRange = feeRange;
   });
   return tableFormatFeesRanges;
 };
+
 
 export default sendYourResponseByEmailController;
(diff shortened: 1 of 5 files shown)


src/main/routes/features/response/eligibility/sendYourResponseByEmailController.ts
```diff
@@ -43,10 +43,23 @@ sendYourResponseByEmailController.get(SEND_RESPONSE_BY_EMAIL_URL, async (req, re
...
-    tableFormatFeesRanges.push(feeRange.formatFeeRangeToTableItem());
+  let previousFeeRange: FeeRange = undefined;
+  feesRanges.sort((element1:FeeRange, element2:FeeRange ) => {
+    if(element1.maxRange < element2.maxRange){
+      return -1;
+    }
+    if(element1.maxRange > element2.maxRange){
+      return 1;
+    }
+    return 0;
```
taomoh, line 54: It would be more readable to extract his logic into a compare() function

taomoh, line 54: It would be more readable to extract his logic into a compare() function

src/main/routes/features/response/eligibility/sendYourResponseByEmailController.ts
```diff
@@ -43,10 +43,23 @@ sendYourResponseByEmailController.get(SEND_RESPONSE_BY_EMAIL_URL, async (req, re
...
+  }).forEach((feeRange: FeeRange) => {
+    if(!feeRange.equals(previousFeeRange)){
+      tableFormatFeesRanges.push(feeRange.formatFeeRangeToTableItem());
+    }
```
taomoh, line 58: I wish there was an equivalent of the following Java one-liner in TS/JS:
`feeRanges.stream().distinct().collect(Collector.toList())` (assumes equals method is implemented adequately in FeeRange class)

@commitBlob @RosDiaf @GorilaNaranja @laperla11 Do you guys know of an equivalent or more succinct way to achieve this?

Please note this is a collection of **objects**. Collection of primary types would work out of the box with this: `const uniqueValues = [...new Set(arrayOfPrimaryTypes)];`

taomoh (commented):
Looks good. Only a couple of comments

src/main/routes/features/response/eligibility/sendYourResponseByEmailController.ts
commitBlob: Maybe [map](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map) can do the trick?
`const uniqueValues = [...new Set(data.map(feeRange=> feeRange.comparedValue))];`

TatsyMaslova: new set did not work. i think because we cannot override equals operator and the values we need are the same but other values are not. even trying to do the same with resulting array does not work because it consists of tableItems. its not a primitive array

TatsyMaslova: i used filter in the end. a bit prettier sort of.

taomoh: I miss Java 😜

laperla11 (approved):

PR #523, opened 2022-06-20
Features/CIV-2909 - Company Statement of Means

commitBlob:
### JIRA link ###
https://tools.hmcts.net/jira/browse/CIV-2909

### Change description ###

- Implement logic to display `statementOfMeans` and `qualifiedStatementOfMeans` depending on the respondent type
- Optionally display hearing requirement details checkbox based on the respondent response type (not displayed if `FULL_ADMISSION`)
- Reset checkboxes selection every time page is refreshed
- Add minor text changes and typo fixes

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/views/macro/types-of-senior-position.njk
+++ b/src/main/views/macro/types-of-senior-position.njk
@@ -10,8 +10,8 @@
    %}
   {% set nonCorporation = '<p class="govuk-body">' + t('COMMON.TYPES_OF_SENIOR_POSITION.CORPORATION_NOT_REGISTERED') + '</p>' %}
   {% set nonCorpPostionsList = '<ul class="govuk-!-padding-left-4"><li>'  +
-    t('COMMON.TYPES_OF_SENIOR_POSITION.NON_CORPORATION_SENIOR_POSITIONS.ANOTHER_OFFICER') + '</li><li>' +
-    t('COMMON.TYPES_OF_SENIOR_POSITION.NON_CORPORATION_SENIOR_POSITIONS.CHAIR')+ '</li></ul>'
+    t('COMMON.TYPES_OF_SENIOR_POSITION.NON_CORPORATION_SENIOR_POSITIONS.CHAIR') + '</li><li>' +
+    t('COMMON.TYPES_OF_SENIOR_POSITION.NON_CORPORATION_SENIOR_POSITIONS.ANOTHER_OFFICER')+ '</li></ul>'
   %}
   {{ govukDetails({
     summaryText: t('COMMON.TYPES_OF_SENIOR_POSITION.TITLE'),


JoshuaLewis99 (approved):

src/main/services/features/response/checkAnswersService.ts
```diff
@@ -429,11 +424,33 @@ export const getSummarySections = (claimId: string, claim: Claim, lang?: string
   return buildSummarySections(claim, claimId, lang);
 };
 
-export const getStatementOfTruth = (claim: Claim) => {
+export const rejectingFullAmount = (claim: Claim): boolean => {
```
TatsyMaslova, line 427: wouldn't it be better if we had that function inside the claim (i.e. claim has the knowledge whether it is rejecting full amount or not). Also our checkAnswersService is getting a bit big and some functions there in my opinion do not belong in that service

src/main/services/features/response/checkAnswersService.ts
```diff
@@ -429,11 +424,33 @@ export const getSummarySections = (claimId: string, claim: Claim, lang?: string
...
   }
-  return new StatementOfTruthForm();
+
+  switch (getSignatureType(claim)) {
+    case SignatureType.BASIC:
+      return new StatementOfTruthForm(rejectingFullAmount(claim));
+    case SignatureType.QUALIFIED:
+      return new QualifiedStatementOfTruth(rejectingFullAmount(claim));
+    default:
+      return new StatementOfTruthForm(rejectingFullAmount(claim));
```
laarmada, line 448: if the default is return new StatementOfTruthForm(rejectingFullAmount(claim)); not make sense have that case: case SignatureType.BASIC:

src/main/views/macro/statement-of-truth.njk
```diff
@@ -1,47 +1,90 @@
 {% from "govuk/components/checkboxes/macro.njk" import govukCheckboxes %}
+{% from "govuk/components/input/macro.njk" import govukInput %}
+{% from "./types-of-senior-position.njk" import typesOfSeniorPosition %}
 
 {% macro statementOfTruth(
   form,
   statementOfTruthType,
-  title = 'Statement of truth',
-  labelHtml = 'I confirm that I believe the details I have provided are correct.',
+  title = t('COMMON.STATEMENT_OF_TRUTH.PAGE_TITLE'),
```
laarmada, line 8: why is in COMMON instead of PAGES.....

src/main/services/features/response/checkAnswersService.ts
commitBlob: There are in total 5 signature types and we are only using Qualified and Basic at the moment. I left switch case if we need to use other types

src/main/views/macro/statement-of-truth.njk
commitBlob: Statement of truth is a component which is injected into check-and-send page, so I should probably change PAGE_TITLE to TITLE?

src/main/services/features/response/checkAnswersService.ts
commitBlob: I can refactor `rejectingFullAmount` into `isFullAmountRejected` and move it to the claim service as it is more generic and can be used elsewhere. I believe other functions are more specific and not sure if they can be used elsewhere, therefore I would like to leave them in and if we use them for something else in the future can move them to more appropriate place.

src/main/services/features/response/checkAnswersService.ts
```diff
@@ -429,11 +424,33 @@ export const getSummarySections = (claimId: string, claim: Claim, lang?: string
   return buildSummarySections(claim, claimId, lang);
 };
 
-export const getStatementOfTruth = (claim: Claim) => {
+export const rejectingFullAmount = (claim: Claim): boolean => {
+  return claim.respondent1.responseType === ResponseType.PART_ADMISSION || claim.respondent1.responseType === ResponseType.FULL_DEFENCE;
+};
+
+export const resetCheckboxFields = (statementOfTruth: StatementOfTruthForm | QualifiedStatementOfTruth): StatementOfTruthForm | QualifiedStatementOfTruth => {
+  statementOfTruth.directionsQuestionnaireSigned = '';
```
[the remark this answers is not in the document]
commitBlob: It is specifically ticket requirement `Reset statement of truth check boxes on 'change'` wouldn't have done it otherwise

src/main/services/features/response/checkAnswersService.ts
laarmada: ok, great then!

src/main/views/macro/statement-of-truth.njk
laarmada: ok I understand, yes Should be change PAGE_TITLE TO TITLE I have done in every screen the PAGE_TITLE is different to TITLE

laarmada (approved):

TatsyMaslova (approved):

PR #549, opened 2022-06-27
Feature/civ 2420 get draft records from legacy draft store

taomoh:
### JIRA link ###

https://tools.hmcts.net/jira/browse/CIV-2420

### Change description ###

Added client for Legacy Draft Service. Part of this client, I had to add a new function inside existing Service Auth Provider client, for generating Service token, as the existing one wasn't fit for purpose: the new function should eventually become the only function used for generating service tokens as it uses the live endpoint `/lease` and one time password mechanism.

The change also includes:
- Integration with Wiremock Legacy Draft Store (configured in civil-sdk, see PR link below), and integration with the actual Service Authorisation Provider
- Legacy draft store URL and secret configuration for test environments

P.S. There is a change to `civil-sdk` that goes hand-in-hand with current change: https://github.com/hmcts/civil-sdk/pull/37  

```
[ ] Yes
[X] No
```


TatsyMaslova (approved):

config/default.yaml
```diff
@@ -24,12 +24,20 @@ services:
...
+        microserviceName: 'cmc'
+        primarySecret: 'not-a-real-primary-secret'
+        secondarySecret: 'not-a-real-secondary-secret'
   dmStore:
     microserviceName: 'xui_webapp'
     baseUrl: 'http://localhost:4506'
   serviceAuthProvider:
     baseUrl: 'http://localhost:4502'
     serviceAuthTokenGeneratorUrl : '/testing-support/lease'
+    url: 'http://localhost:4502'
```
laperla11, line 39: we already have this url on line 37

src/main/routes/features/dashboard/dashboardController.ts
```diff
@@ -25,14 +26,19 @@ function renderPage(res: express.Response, claimsAsClaimant: Claim[], claimDraft
 
 const dashboardController = express.Router();
 
-dashboardController.get(DASHBOARD_URL, async function (req, res) {
-  const claimsAsClaimant : Claim[] = [];
-  const claimsAsDefendant : CivilClaimResponse[]  = await civilServiceClient.retrieveByDefendantId(<AppRequest>req);
+dashboardController.get(DASHBOARD_URL, async function (req: AppRequest, res) {
+  const user: UserDetails = req.session.user;
+  // This is a call to validate integration with legacy draft-store. This will have to be refined in the future
```
commitBlob, line 31: for block comments you can use `/* */`

commitBlob (approved):

laperla11 (approved):

src/main/app/client/legacyDraftStoreClient.ts
```diff
@@ -0,0 +1,44 @@
...
+const cmcS2sSecret = config.get<string>('services.serviceAuthProvider.cmcS2sSecret');
+const primarySecret = config.get<string>('services.draftStore.legacy.s2s.primarySecret');
+const secondarySecret = config.get<string>('services.draftStore.legacy.s2s.secondarySecret');
+const microserviceName = config.get<string>('services.draftStore.legacy.s2s.microserviceName');
+
+const secretsAsHeader = (): string => {
+  return secondarySecret ? `${primarySecret}, ${secondarySecret}` : primarySecret;
+};
+
+const find = async (userToken: string): Promise<void> => {
```
laperla11, line 19: I wouldn't use `find` as function name as it is built-in JS method, maybe use `findUser` instead?

GorilaNaranja (approved):

src/main/routes/features/dashboard/dashboardController.ts
taomoh: Incorporated

config/default.yaml
taomoh: Incorporated

src/main/app/client/legacyDraftStoreClient.ts
taomoh: Good point. Incorporated

PR #582, opened 2022-07-05
Civ 2988 welsh translation error messages

laarmada:
### JIRA link ###

https://tools.hmcts.net/jira/browse/CIV-2988

### Change description ###

Develop the CCUI framework for handling translation of error messages and their display in the Welsh language

```
[ ] Yes
[x] No
```


src/main/modules/nunjucks/index.ts
```diff
@@ -65,13 +66,21 @@ export class Nunjucks {
 
     const currencyFormat = (value: number) => numeral.default(value);
 
+    const errorSummaryTranslation = (keys: FormValidationError[], t: any) => {
+      for (let i = 0; i < keys.length; i++) {
+        keys[i].text = t(keys[i].text);
```
taomoh, line 71: It would be safer to use a separate variable for the return, instead of re-using the parameter, and name it appropriately, e.g. translatedErrors

src/main/modules/nunjucks/index.ts
```diff
@@ -65,13 +66,21 @@ export class Nunjucks {
...
+    };
+
     nunjucksEnv.addGlobal('asset_paths', appAssetPaths);
     nunjucksEnv.addGlobal('development', this.developmentMode);
     nunjucksEnv.addGlobal('govuk_template_version', packageDotJson.dependencies.govuk_template_jinja);
     nunjucksEnv.addFilter('currencyFormat', currencyFormat);
     nunjucksEnv.addFilter('addDays', addDaysFilter);
     nunjucksEnv.addFilter('date', dateFilter);
     nunjucksEnv.addGlobal('t', (key: string, options?: TOptions): string => this.i18next.t(key, options));
+    nunjucksEnv.addGlobal('errorSummaryTranslation', errorSummaryTranslation);
```
taomoh, line 83: There is no view in this PR using this new function. I assume you've tested locally but didn't commit the change in the view and you left it to future stories to incorporate the call to the new function inside the njk views?

taomoh (changes requested):

src/main/modules/nunjucks/index.ts
taomoh: Also better to call the function in a verbal mode, e.g. translateErrors()

laarmada: have a look @taomoh I have make changes.

taomoh: Looking good. Thanks @laarmada

taomoh (approved):

PR #599, opened 2022-07-12
Civ 2824 client to submit event

TatsyMaslova:

### JIRA link ###
https://tools.hmcts.net/jira/browse/CIV-2824

### Change description ###

```
[ ] Yes
[x ] No
```

### DIFF
--- a/src/main/app/client/civilServiceClient.ts
+++ b/src/main/app/client/civilServiceClient.ts
@@ -7,7 +7,7 @@ import {
   CIVIL_SERVICE_CASES_URL,
   CIVIL_SERVICE_DOWNLOAD_DOCUMENT_URL,
   CIVIL_SERVICE_FEES_RANGES,
-  CIVIL_SERVICE_SUBMIT_RESPONSE_EVENT_TOKEN,
+  CIVIL_SERVICE_SUBMIT_EVENT,
 } from './civilServiceUrls';
 import {FeeRange, FeeRanges} from '../../common/models/feeRange';
 import {plainToInstance} from 'class-transformer';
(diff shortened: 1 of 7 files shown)


src/main/app/client/civilServiceClient.ts
```diff
@@ -127,15 +126,20 @@ export class CivilServiceClient {
     }
   }
 
-  async getSubmitDefendantResponseEventToken(claimId: string, req: AppRequest): Promise<string> {
+  async submitDefendantResponseEvent(claimId: string, req: AppRequest): Promise<Claim> {
```
taomoh, line 129: It might be worth making this function generic, by renaming it to submitEvent and adding an event argument. So it can be re-used by future event submissions.

src/main/common/models/events/eventDto.ts
```diff
@@ -0,0 +1,7 @@
+import {CaseEvent} from './caseEvent';
+
```
osheaf01, line 2: In the same vein as Taoufik's comment above about generics, you might make this EventDto an interface, and have a CaseEventDto that implements it?

osheaf01 (approved):
One small suggestion

src/main/common/models/events/eventDto.ts
TatsyMaslova: CaseEvent is an enum. EventDto is an interface as you can see. but its not an interface like in java its something to represent a json that is returned. basically it is a data transfer object that we are passing to civil service. civil service expects that dto if you have a look at the event dto on the civil service side

src/main/app/client/civilServiceClient.ts
TatsyMaslova: extracted a method that takes event as a paramether this one is using it to  send specific event to submit a response event

PR #603, opened 2022-07-14
CIV-915 - Response Deadline Options

commitBlob:
### JIRA link ###
https://tools.hmcts.net/jira/browse/CIV-915

### Change description ###
- implement response deadline option page
- update task list item - need more time with task status complete logic depending on the response deadline option selection

```
[ ] Yes
[x] No
```


TatsyMaslova (approved):

src/main/routes/features/response/responseDeadlineOptionsController.ts
```diff
@@ -0,0 +1,81 @@
...
+  });
+
+responseDeadlineOptionsController.post(RESPONSE_DEADLINE_OPTIONS_URL,
+  async (req: express.Request, res: express.Response, next: express.NextFunction) => {
+    try {
+      let responseOption;
+      let redirectUrl;
+      const claimId = req.params.id;
+      const claim = await getCaseDataFromStore(claimId);
+      switch (req.body['option']) {
```
laarmada, line 47: only personal opinion is so large the controller that switch should be in other method.

laarmada (approved):

laperla11 (approved):

src/main/routes/features/response/responseDeadlineOptionsController.ts
commitBlob: I did think about extracting switch into service function but I don't like the way JavaScript deals with multiple return params, therefore I left it as a part of the controller. Alternative would be having separate service function for each param but that way we're duplicating switch logic.
In JavaScript in order to return multiple params you would need to put them in array `return [returnValue1, returnValue2]` and then you access them by index.
Contrary to Python (and probably some other languages) you can do:
`return returnValue1, returnValue2` which would be in controller: `controllerValue1, controllerValue2 = functionWithMultipleReturnParams(options)`

PR #610, opened 2022-07-18
CIV-3118 - Request More Time

commitBlob:
### JIRA link ###
https://tools.hmcts.net/jira/browse/CIV-3118

### Change description ###
- implement request more time page
- add logic to mark need more time task as complete if more then 28 days is requested to respond to the claim

```
[ ] Yes
[x] No
```


TatsyMaslova (approved):

src/main/services/features/response/responseDeadlineService.ts
```diff
@@ -1,16 +1,49 @@
 import {getCaseDataFromStore, saveDraftClaim} from '../../../modules/draft-store/draftStoreService';
 import {ResponseDeadline, ResponseOptions} from '../../../common/form/models/responseDeadline';
+import {AdditionalTimeOptions} from '../../../common/form/models/additionalTime';
 
 const {Logger} = require('@hmcts/nodejs-logging');
 const logger = Logger.getLogger('partialAdmissionService');
 
 export class ResponseDeadlineService {
+  getAdditionalTime(additionalTime: string): AdditionalTimeOptions {
+    switch (additionalTime) {
```
laarmada, line 10: we can make the same in one line like that:
  return additionalTime===AdditionalTimeOptions.UP_TO_28_DAYS ? AdditionalTimeOptions.UP_TO_28_DAYS
    : AdditionalTimeOptions.MORE_THAN_28_DAYS;

laarmada (changes requested):

src/main/services/features/response/responseDeadlineService.ts
commitBlob: But that won't work if `additionalTime = 'something-unlikely' ` as it would return `AdditionalTimeOptions.MORE_THAN_28_DAYS` for that case

laarmada: great call!!! Thanks for your time

laarmada (approved):

src/main/services/features/response/responseDeadlineService.ts
commitBlob: Pushed up new changes on this

PR #714, opened 2022-08-08
CIV-3826 - Log out functionality

commitBlob:
### JIRA link ###
https://tools.hmcts.net/jira/browse/CIV-3826

### Change description ###
- update header component to show appropriate links
- terminate session both locally and with idam
- refactor header component into reusable macro

```
[] Yes
[] No
```


src/main/views/question-template.njk
```diff
@@ -1,11 +1,10 @@
-{% from "govuk/components/back-link/macro.njk" import govukBackLink %}
```
GorilaNaranja, line 1: Just for curiosity, do you know why we extend from _question-template.njk_ instead _claim-details-tpl.njk_ only in priority-debts.njk page? 

Maybe we can delete this file and in priority-debts extend from claim-details-tpl

commitBlob: I am not sure why we've have couple of templates, but question template also contains errorSummary wrapper which is different compared to other pages which use claim-details template. 
I created tech debt ticket for this already as I didn't want to tackle it within the scope of this ticket.
https://tools.hmcts.net/jira/browse/CIV-4039

RosDiaf (approved):

commitBlob:
closing as idam does not seem to work on this pr

PR #724, opened 2022-08-11
Bug/civ 3908 repayment schedule translation

RosDiaf:

### Change description ###

```
[ ] Yes
[ ] No
```


TatsyMaslova (approved):

src/main/assets/js/calculate-length-repayment.js
```diff
@@ -38,29 +57,34 @@
...
+      const toggleSchedule = (containerOne, containerTwo, containerThree) => {
+        numberOfInstalments === 1 ? containerOne.classList.remove(CLASS_HIDE) : containerOne.classList.add(CLASS_HIDE);
+        numberOfInstalments === 2 ? containerTwo.classList.remove(CLASS_HIDE) : containerTwo.classList.add(CLASS_HIDE);
+        numberOfInstalments > 2 ? containerThree.classList.remove(CLASS_HIDE) : containerThree.classList.add(CLASS_HIDE);
+      };
+
       switch (val) {
         case 'WEEK':
-          repaymentLength.innerHTML = numberOfInstalments ? numberOfInstalments + pluralize(numberOfInstalments, 'week') : '-';
+          hideAll();
```
hmcts-version1-pablo, line 70: I think this is a very good solution Rosario. I was the one who was struggling with this and as we were running out of time I thought it best to leave it in a bug ticket and I couldn't have done it as well as you.

A suggestion that I don't know if it would be possible. Instead of doing inside each switch hideAll , could be done only 1 time before the case and avoid repeating code no? Regarding getNumberOfInstalments(), I don't know if you could do something similar, although I understand that in TWO_WEEKS there is an exceptional case for this call.

For the rest, I see it perfect! Very good work!

RosDiaf: You're right. I have just moved the hideAll method outside the switch case.

hmcts-version1-pablo (approved):

PR #729, opened 2022-08-12
Civ 3737 initiate response event

TatsyMaslova:

### JIRA link ###
https://tools.hmcts.net/jira/browse/CIV-3737

### Change description ###

```
[ ] Yes
[ ] No
```


src/main/app/client/civilServiceClient.ts
```diff
@@ -133,16 +133,21 @@ export class CivilServiceClient {
...
-  async submitEvent(event: CaseEvent, claimId: string, req: AppRequest): Promise<Claim> {
+  async submitEvent(event: CaseEvent, claimId: string, updatedClaim?:ClaimUpdate, req?: AppRequest): Promise<Claim> {
     const config = this.getConfig(req);
     const userId = req.session?.user?.id;
     const data : EventDto = {
       event:event,
-      caseDataUpdate: new Map<string, string>(),
+      caseDataUpdate: updatedClaim,
     };
+    console.log(data);
```
commitBlob, line 150: Is this just for testing purposes?
Maybe we can use logger saying something like "initiated submit event + data"? With that we could keep in permanently in the application rather than going back some time later and removing console logs.

src/main/common/models/claim.ts
```diff
@@ -214,6 +217,23 @@ export class Claim {
   isDefendantNotResponded(): boolean {
     return this.ccdState === CaseState.AWAITING_RESPONDENT_ACKNOWLEDGEMENT;
   }
+
+  isDeadlineExtended(): boolean {
+    return this.responseDeadline?.option === ResponseOptions.ALREADY_AGREED && this.respondentSolicitor1AgreedDeadlineExtension !== undefined;
```
commitBlob, line 222: I don't think you need undefined check here `this.respondentSolicitor1AgreedDeadlineExtension !== undefined`
undefined

commitBlob: Can you please check task list logic for displaying request more time step?
Believe code you added might have potential to implement in that step. If that is the case can you raise tech debt please

commitBlob (approved):

src/main/common/models/claim.ts
TatsyMaslova: yes, task list logic where it idetifies whether the task for extension is complete has the same logic but differently implemented with switch case. i will raise a tech debt ticket so that it reuses the method in the claim. Then when the field changes (which i think it will) for the extending response for citizen (instead of solicitor) we will have to change the check only in one place (instead of guard, task list, etc)

GorilaNaranja (approved):

PR #755, opened 2022-08-23
CIV-3738: Do you have a postal address in the UK?

hmcts-version1-pablo:
### JIRA link ###

https://tools.hmcts.net/jira/browse/CIV-3738

### Change description ###
Do you have a postal address in the UK? screen

```
[ ] Yes
[X ] No
```


src/main/modules/i18n/locales/cy.json
```diff
@@ -560,6 +560,11 @@
       "PAGE_TITLE": "A oes gennych anabledd?",
       "TITLE": "A oes gennych anabledd?"
     },
+    "ELIGIBILITY_CLAIMANT_ADDRESS" : {
+      "PAGE_TITLE": "Do you have a postal address in the UK?",
```
laperla11, line 564: both TITLE and PAGE_TITLE have the same content, one if them is unnecessary for translation.

laperla11 (approved):

src/main/modules/i18n/locales/cy.json
hmcts-version1-pablo: I have put 2 keys in the translations for when they are translated into Welsh. At least according to Sergio, the PAGE_TITLE are not translated, so even if they are the same we should have 2.

hmcts-version1-domini (approved):

TatsyMaslova (approved):

PR #777, opened 2022-08-26
CIV-3764 cant use this service 18 or over

hmcts-version1-domini:
### JIRA link ###

https://tools.hmcts.net/jira/browse/CIV-3764

### Change description ###

Develop the 'You can't use this service' - You need to be 18 or over to use this service.

```
[ ] Yes
[x] No
```


src/main/common/form/models/eligibility/NotEligibleReason.ts
```diff
@@ -4,6 +4,7 @@ export enum NotEligibleReason {
   CLAIM_VALUE_OVER_25000 = 'claim-value-over-25000',
   UNDER_18 = 'under-18',
   UNDER_18_DEFENDANT = 'under-18-defendant',
+  UNDER_18_CLAIMANT = 'under-18',
```
portizdev, line 7: I think you have duplicated the constant, in line 5 you have 'under-18' I think it is the same

src/main/common/utils/notEligibleReasonConvertor.ts
```diff
@@ -9,7 +9,7 @@ export const convertToNotEligibleReason = (reason: string): NotEligibleReason =>
     case 'claim-value-over-25000':
       return NotEligibleReason.CLAIM_VALUE_OVER_25000;
     case 'under-18':
-      return NotEligibleReason.UNDER_18;
+      return NotEligibleReason.UNDER_18_CLAIMANT;
```
portizdev, line 12: Why this change?

portizdev (changes requested):

src/main/common/form/models/eligibility/NotEligibleReason.ts
hmcts-version1-domini: Ok. Removed the generic one as we have talked about.

src/main/common/utils/notEligibleReasonConvertor.ts
hmcts-version1-domini: because I wanted the constants to have the name with the same structure as the UNDER 18 DEFENDANT

portizdev (approved):

PR #846, opened 2022-09-14
CIV-4658 - Features - DQ - Tried to Settle the Claim

commitBlob:
### JIRA link ###
https://tools.hmcts.net/jira/browse/CIV-4658

### Change description ###
- implement tried to settle the claim and appropriate unit tests
- fix couple of typos
- re-order alphabetically test messages

```
[ ] Yes
[x] No
```


RosDiaf (approved):

laperla11 (approved):

TatsyMaslova (approved):

src/main/common/models/claim.ts
```diff
@@ -77,6 +78,7 @@ export class Claim {
   ccdState: CaseState;
   responseDeadline: ResponseDeadline;
   determinationWithoutHearing: DeterminationWithoutHearing;
+  directionQuestionnaire?: DirectionQuestionnaire;
```
laarmada, line 81: if you have add directionQuestionnaire?: DirectionQuestionnaire; to the model, maybe we need to include the other related as:
vulnerability
determinationWithoutHearing
..... and the next ones....

at this way we could use only one service for all cases instead of one for each, saving time, effort and code.
what do you think?

laarmada (changes requested):

src/main/common/models/claim.ts
commitBlob: I believe it will be rather big one service, that's why I've separated it, but agree that all dq related fields should be part of the dq model, we are waiting sampan's ticket to get the list of all fields

laarmada: I have done the dynamic service in my branch is so easy to do that, we can see after the stand up.

laarmada (approved):

PR #852, opened 2022-09-15
CIV-4666 - Features - DQ - Shared Expert

commitBlob:
### JIRA link ###
https://tools.hmcts.net/jira/browse/CIV-4666

### Change description ###
Implement shared expert page and appropriate tests

```
[ ] Yes
[x] No
```


src/main/routes/features/directionsQuestionnaire/sharedExpertController.ts
```diff
@@ -0,0 +1,46 @@
...
+  getSharedExpertForm,
+  getSharedExpertSelection,
+  saveSharedExpertSelection,
+} from '../../../services/features/directionsQuestionnaire/sharedExpertService';
+
+const sharedExpertController = express.Router();
+
+function renderView(sharedExpertForm: GenericForm<GenericYesNo>, res: express.Response): void {
+  const form = Object.assign(sharedExpertForm);
+  form.option = sharedExpertForm.model.option;
```
laarmada, line 16: why you dont use sharedExpertForm directly and use inside of .njk sharedExpertForm.model.option? we can save create a new object "form"...

laarmada (approved):

RosDiaf (approved):

src/main/routes/features/directionsQuestionnaire/sharedExpertController.ts
commitBlob: If you don't do this yes no nunjucks will not capture your selection when you come back to the page because generic form has option property nested within model rather than being flatten.
nunjucks yesno expects `{option: 'yes/no}`, and if we don't do above it will receive `{model: {option: 'yes/no}}`,

Not sure why are you flagging this on this pr as it is not anything new, we've been dealing with it across all screens.

laarmada: because I dont need to do that for recover the data:

commitBlob: I see what is happening, I am using yesNo macro component instead of genericYesNo.
I changed that in the #856 as Raja was not able to sign into environment with this PR

commitBlob:
closing as #856 is used instead

PR #854, opened 2022-09-16
Civ 4663 dq expert evidence

laarmada:
### JIRA link ###

https://tools.hmcts.net/jira/browse/CIV-4663

### Change description ###

 the development of the 'Do you want to use expert evidence' screen.

```
[x] Yes
[ ] No
```


src/main/common/models/directionsQuestionnaire/defendantExpertEvidence.ts
```diff
@@ -0,0 +1,11 @@
+import {YesNo} from '../../form/models/yesNo';
+import {IsDefined} from 'class-validator';
+
+export class DefendantExpertEvidence {
```
TatsyMaslova, line 4: have we agreed on how we are going to do this? seems that everything else is using generic yesno model. would be weird if we have one thing one way and one another..

src/main/common/models/directionsQuestionnaire/defendantExpertEvidence.ts
```diff
@@ -0,0 +1,11 @@
...
+  @IsDefined({message: 'ERRORS.DEFENDANT_EXPERT_EVIDENCE_REQUIRED'})
+    option?: YesNo;
+
+  constructor(option?: YesNo, details?: string) {
```
commitBlob, line 8: where are details coming from?
how different this model is from generic yes/no one?

src/main/modules/i18n/locales/en.json
```diff
@@ -677,6 +677,14 @@
       "ADD_ANOTHER_EVENT": "Add another event",
       "ADD_ANY_COMMENTS": "Add any comments about their timeline (optional)"
     },
+    "DEFENDANT_EXPERT_EVIDENCE": {
```
commitBlob, line 680: what happened to welsh keys?

src/main/services/features/directionsQuestionnaire/directionQuestionnaireService.ts
```diff
@@ -0,0 +1,38 @@
...
+
+const {Logger} = require('@hmcts/nodejs-logging');
+const logger = Logger.getLogger('directionQuestionnaireService');
+
+const getDirectionQuestionnaire = async (claimId: string): Promise<DirectionQuestionnaire> => {
+  try {
+    const claim = await getCaseDataFromStore(claimId);
+    if (!claim.directionQuestionnaire) return new DirectionQuestionnaire();
+    return new DirectionQuestionnaire(
+      claim.directionQuestionnaire?.triedToSettle,
```
commitBlob, line 12: whenever we add new DQ property we need to keep in mind to add it here?

src/main/views/features/directionsQuestionnaire/defendant-expert-evidence.njk
```diff
@@ -0,0 +1,46 @@
+{% extends "claim-details-tpl.njk" %}
+{% from "govuk/components/button/macro.njk" import govukButton %}
+{% from "govuk/components/inset-text/macro.njk" import govukInsetText %}
```
commitBlob, line 3: you are not using govukInsetText  and govukTextarea

src/main/common/models/directionsQuestionnaire/defendantExpertEvidence.ts
laarmada: is because el message is different I will use your approach here. let me change

src/main/services/features/directionsQuestionnaire/directionQuestionnaireService.ts
laarmada: if you use that get in your controller else not

src/main/views/features/directionsQuestionnaire/defendant-expert-evidence.njk
laarmada: good point... copy and paste :D

src/main/common/models/directionsQuestionnaire/defendantExpertEvidence.ts
laarmada: yes I just to remove it.

src/main/common/models/directionsQuestionnaire/defendantExpertEvidence.ts
laarmada: fix in the last commit. removed

src/main/modules/i18n/locales/en.json
laarmada: fix in the last commit.

portizdev (approved):

src/main/services/features/directionsQuestionnaire/directionQuestionnaireService.ts
commitBlob: But we have at the moment 
`triedToSettle, defendantExpertEvidence, sharedExpert, requestExtra4weeks` properties in the direction questionnaire.
What you are doing here is setting only `triedToSettle` and `defendantExpertEvidence,`.
So if `sharedExpert` or `requestExtra4weeks` were already set in the redis you are resetting it with this new DirectionQuestionnaire as you are not passing those values via constructor yes?
We will add more than these 4 properties and and whenever we add a new one we need to keep in mind that we need to include it here, otherwise values will be always reset when we use this getDirectionQuestionnaire. 
For me this seems like shooting ourselves in the foot.

commitBlob: after the team demo as we discussed, believe we wont need `return new DirectionQuestionnaire(...)` instead we will return `return claim.directionQuestionnaire)`

laarmada: commit that change

commitBlob (approved):

PR #868, opened 2022-09-21
CIV-5005 - TechDebt - DQ Services Refactor

commitBlob:
### JIRA link ###
https://tools.hmcts.net/jira/browse/CIV-5005

### Change description ###
- refactor DQ views which were using yesNoForm macro into yesNoGeneric macro (with this we don't need to do object.assing on the renderView for most of the pages)
- refactor properties which were relevant to the DQ into the DQ model
- use generic DQ service where applicable (cannot be done on all properties)

**Note:**
- was not able to refactor DQ error messages into enum as dynamic setting (dqErrorMsg[dynamicName]) is not TS compliant
- code smell introduced on DQ model (Constructor has too many parameters (9). Maximum allowed is 7) don't think we can do anything about it

```
[ ] Yes
[x] No
```


laarmada (approved):
really good save and optimization of code!

src/main/routes/features/directionsQuestionnaire/requestExtra4WeeksController.ts
```diff
@@ -2,24 +2,24 @@ import * as express from 'express';
...
-} from '../../../services/features/directionsQuestionnaire/requestExtra4WeeksService';
 import {constructResponseUrlWithIdParams} from '../../../common/utils/urlFormatter';
+import {
+  getGenericOption,
+  getGenericOptionForm,
+  saveDirectionQuestionnaire,
+} from '../../../services/features/directionsQuestionnaire/directionQuestionnaireService';
 
 const requestExtra4WeeksController = express.Router();
+const requestExtra4weeksErrorMessage = 'ERRORS.VALID_REQUEST_EXTRA_4_WEEKS';
```
TatsyMaslova, line 13: shame we are not doing the enum for error messages and the controller has to know about them

commitBlob: Yeah, was not able to do it dynamically  😞 
This `errorMessagesEnum[dynamicName]` is not compliant with TS, I tried to do it hacky way `${errorMessagesEnum}.${dynamicName}` but this would display as `[object object].dynamicName`

commitBlob: Should I try to do initially proposed switch, or keep it as it is?

commitBlob: I will try to do it as a constant now

TatsyMaslova (approved):
looks great. i did mention about how we possibly could do it whtouht as keyof typeof. but maybe not worth the effort in this case.

PR #890, opened 2022-09-29
CIV-4226 - Feature - Claimant Date of Birth

commitBlob:
### JIRA link ###
https://tools.hmcts.net/jira/browse/CIV-4226

### Change description ###
Implement claimant date of birth page

```
[ ] Yes
[x] No
```


RosDiaf (approved):

src/main/views/features/public/claim/claimant-dob.njk
```diff
@@ -0,0 +1,86 @@
+{% extends "claim-details-tpl.njk" %}
```
laperla11, line 1: isn't it possible to use `citizen-dob.njk`?

laperla11 (changes requested):

src/main/views/features/public/claim/claimant-dob.njk
commitBlob: Seems like title is slightly different, will see if I can try to come up with something smart.

PR #921, opened 2022-10-03
CIV-4683: Defendant Response - DQ - Why are you, your experts or witnesses unavailable for a hearing for <<xx>> days?

portizdev:
### JIRA link ###
https://tools.hmcts.net/jira/browse/CIV-4683

### Change description ###
- "Why are you, your experts or witnesses unavailable for a hearing for <<xx>> days?" screen
- TODO: Change the hardcoded 22 to the correct number of days (To be calculated on the page "Are there any dates in the next 12 months when you, your experts or your witnesses cannot attend a hearing?")

```
[ ] Yes
[X ] No
```


src/main/modules/i18n/locales/cy.json
```diff
@@ -1498,6 +1498,11 @@
       "PAGE_TITLE": "Your employer",
       "WHO_EMPLOYS": "Pwy sy’n eich cyflogi"
     },
+    "WHY_UNAVAILABLE_FOR_HEARING": {
```
laarmada, line 1501: is not at the same line of en.json file

laarmada (changes requested):

src/main/modules/i18n/locales/cy.json
portizdev: A translation of the response submission pages was missing. Fixed.

src/main/routes/features/directionsQuestionnaire/whyUnavailableForHearingController.ts
```diff
@@ -0,0 +1,54 @@
...
+
+const whyUnavailableForHearingController = express.Router();
+const whyUnavailableForHearingViewPath = 'features/directionsQuestionnaire/why-unavailable-for-hearing';
+const dqPropertyName = 'whyUnavailableForHearing';
+const dqParentName = 'hearing';
+
+function renderView(form: GenericForm<WhyUnavailableForHearing>, res: express.Response): void {
+  //TO DO: Change the hardcoded 22 to the correct number of days (To be calculated on the page "Are there any dates in
+  // the next 12 months when you, your experts or your witnesses cannot attend a hearing?")
+  res.render(whyUnavailableForHearingViewPath, {form, days:22});
```
commitBlob, line 22: this might be hard todo to spot and it would require a bit of refactoring as someone would need to change renderView, get, post, and create calculation service function.
I would pass the days to renderView as a param which would come from the service call in the get and post controllers.

portizdev: Ok, I will create the service (without the calculation) and create the call in get and post and the parameter in render View.
I leave the TODO only in the service.

laperla11: I would recommend also creating a tech debt ticket to monitor this TODO

laperla11 (approved):

src/main/routes/features/directionsQuestionnaire/whyUnavailableForHearingController.ts
portizdev: Done! https://tools.hmcts.net/jira/browse/CIV-5151

commitBlob (approved):

laarmada (approved):

PR #930, opened 2022-10-05
Increase Pipeline Tests Performance

commitBlob:
### JIRA link ###
n/a

### Change description ###
Add `-i` (run in band) flag which should increase the performance of test in the CPU constrained environments

https://stackoverflow.com/questions/43864793/why-does-jest-runinband-speed-up-tests
https://github.com/facebook/jest/issues/2586
https://itnext.io/how-to-make-your-sluggish-jest-v23-tests-go-faster-1d4f3388bcdd

```
[ ] Yes
[x] No
```


commitBlob:
Closing as pipeline is running out of memory when running one process node

PR #936, opened 2022-10-06
Add back in the cmc-s2s-secret

chris-pateman:
### Change description ###
Add back in the cmc-s2s-secret as required for other environments

```
[ ] Yes
[ x ] No
```


chris-pateman:
not required

PR #983, opened 2022-10-14
CIV-4339 - Features - Total Claim Interest

commitBlob:
### JIRA link ###
https://tools.hmcts.net/jira/browse/CIV-4339

### Change description ###
Implement 'What is the total interest for your claim` view
Implement Help calculating interest macro component

```
[ ] Yes
[x] No
```


src/main/views/features/claim/interest/total-claim-interest.njk
```diff
@@ -0,0 +1,75 @@
...
+      </div>
+      <div class="govuk-grid-row">
+        <div class="govuk-grid-column-two-thirds">
+          <h1 class="govuk-heading-l">{{ t('PAGES.TOTAL_INTEREST.TITLE') }}</h1>
+          <p class="govuk-body">
+            {{ t('PAGES.TOTAL_INTEREST.CALCULATE_INTEREST') }}
+            {{ t('PAGES.TOTAL_INTEREST.ADD_TO_CLAIM_AMOUNT') }}
+          </p>
+
+          {{ helpCalculatingInterest(t) }}
```
laperla11, line 31: I prefer not to use macro for static content, we can just import the static content with `include` so we don't need to pass `t` function

laperla11 (approved):

src/main/views/macro/contact-us.njk
```diff
@@ -17,7 +13,7 @@
   <p class="govuk-body govuk-!-margin-bottom-0">{{responseDelay}}</p>
   <h3 class="govuk-heading-m govuk-!-margin-bottom-0">Telephone</h3>
   <p class="govuk-body govuk-!-margin-bottom-0">{{ phone }}</p>
-  <p class="govuk-body govuk-!-margin-bottom-0">{{openingHours}}</p>
+  <p class="govuk-body govuk-!-margin-bottom-0">{{openingHours}}
```
laarmada, line 16: take care that extra line, that component is used in a lot of screens. I trust you :D

laarmada (approved):
Could you please improve the coverage? https://sonarcloud.io/summary/new_code?id=civil-citizen-ui&pullRequest=981

src/main/views/macro/contact-us.njk
commitBlob: I removed it because here `<a target="_blank" class="govuk-link" href={{callChargesUrl}}>{{callCharges}}</a></p>`
P tag is closed, so we had one more extra closure

src/main/views/features/claim/interest/total-claim-interest.njk
commitBlob: I see your point. Agree and disagree 😅 
As it is static content I can do include instead, however if I do so I need to move this outside of macro directory and I am afraid if it is not there someone might create their own. Since there is AC to create reusable component and I am not sure if the text on other screen(s) is going to be 100% I would like to leave it as a macro and if someone needs to tweak the content they can pass it via function and add parameter and wrap content in if/else

PR #991, opened 2022-10-18
Tech debt/civ 5040 regular expenses missing translations

RosDiaf:

### Change description ###

```
[ ] Yes
[ ] No
```


src/main/modules/i18n/locales/en.json
```diff
@@ -1735,6 +1746,55 @@
     "ENTER_VALID_SECURITY_CODE": "Enter a valid security code",
     "ENTER_VALID_TITLE": "Enter valid title",
     "EQUAL_INSTALMENTS_REQUIRED": "Enter a valid amount for equal instalments",
+    "EXPENSES_ENTER_OTHER_SOURCE": "Enter other expense source",
+    "EXPENSES_AMOUNT": {
+      "COUNCIL_TAX": "Enter how much you pay for council tax",
```
commitBlob, line 1751: These look like something which can be generic `Enter how much you pay for {{EXPENSE}}`.
Can we not do something like this?

src/main/modules/i18n/locales/en.json
```diff
@@ -16,6 +16,20 @@
     "DATE_EXAMPLE": "For Example, {{ date }}",
     "DAY": "Day",
     "DOWNLOAD_THE_CLAIM": "Download the claim",
+    "EXPENSE_COUNCIL_TAX": "Council Tax",
```
commitBlob, line 19: I think this can be even more common by not being an expense believe we are using same values for debts

RosDiaf: I moved expenses under CHECKBOX_FIELDS together with debt so they both share the same section.

portizdev (approved):

commitBlob (approved):

PR #994, opened 2022-10-19
Update the redis data for testing 5144

laperla11:
### Change description ###
- Update the redis data for testing 5144

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/modules/draft-store/redisData.json
+++ b/src/main/modules/draft-store/redisData.json
@@ -122,7 +122,7 @@
           "AddressLine2": "Caversham House 15-17",
           "AddressLine3": "Church Road"
         },
-        "telephoneNumber": "01632960001"
+        "phoneNumber": "01632960001"
       },
       "claimInterest": "Yes",
       "submittedDate": "2022-03-01T11:57:10.592492",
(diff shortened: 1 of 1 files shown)


laarmada (approved):

GorilaNaranja (approved):

src/main/modules/draft-store/redisData.json
```diff
@@ -450,7 +450,7 @@
       "issueDate": "2022-01-08",
       "applicant1": {
         "type": "INDIVIDUAL",
-        "telephoneNumber": "(0)207 127 0000",
+        "phoneNumber": "(0)207 127 0000",
```
TatsyMaslova, line 453: Kadir, in ccd this is partyPhone field... I know that as a result if we wanted to retrieve the phone from the lr claim in the release1 it would have never worked because we mocked the field wrong initially. but should we change it to correspond the ccd (since we don't have the translation yet?) then when we do test it with real data we can actually see the phone?

laperla11: changed to partyPhone for now until translation is completed.

TatsyMaslova (approved):

PR #996, opened 2022-10-19
Fix: mediationPhoneNumber of undefined

GorilaNaranja:
### JIRA link ###
https://tools.hmcts.net/jira/browse/CIV-5140

### Change description ###
CUI : Full defence response with No Mediation option is not working

```
[ ] Yes
[x] No
```


commitBlob (approved):
Would be great if we had unit test for such case

laperla11 (approved):

PR #1029, opened 2022-10-28
CIV-3144 - Tech Debt - Resetting redis data with changing response type.

neeta-hmcts:
### JIRA link ###
https://tools.hmcts.net/jira/browse/CIV-3144

### Change description ###
Added a code to reset the redis data when user changes the response type.

```
[ ] Yes
[x] No
```


src/main/routes/features/response/responseType/citizenResponseTypeController.ts
```diff
@@ -52,6 +55,8 @@ citizenResponseTypeController.post(CITIZEN_RESPONSE_TYPE_URL,
           claim.respondent1 = respondent;
         }
         await saveDraftClaim(req.params.id, claim);
+        await resetPreviousResponseTypeSettings(req.params.id);
```
commitBlob, line 58: I think this call should happen before `saveDraftClaim` otherwise we are overwriting selections we just made (line 57) ?

commitBlob (commented):
I think this should be demoed to the whole team as it would be easier to visualise the change

commitBlob:
removing labels until it passes dev review

src/main/services/features/response/resetPreviousResponseTypeSettingService.ts
```diff
@@ -0,0 +1,39 @@
...
+    if (claim.rejectAllOfClaim) {
+      claim.rejectAllOfClaim = new RejectAllOfClaim();
+    }
+
+    if (claim.statementOfMeans) {
+      claim.statementOfMeans = new StatementOfMeans();
+    }
+
+    //clear common fields
+    claim.paymentOption = undefined;
```
laarmada, line 26: we need to move to the correspondent hierarchy no make sense have in claim....

src/main/routes/features/response/responseType/citizenResponseTypeController.ts
laarmada, line 58: I don't have clear if that is correct, you will remove everything under:
 claim.partialAdmission
 claim.rejectAllOfClaim
claim.statementOfMeans

after that page.... so we will loose information

laarmada (changes requested):

src/main/routes/features/response/responseType/citizenResponseTypeController.ts
laarmada: take care with that if we remove statementOfMeans before to save maybe we loose data from the previous steps....

laarmada (changes requested):

src/main/services/features/response/resetPreviousResponseTypeSettingService.ts
```diff
@@ -0,0 +1,39 @@
...
+
+    if (claim.rejectAllOfClaim) {
+      claim.rejectAllOfClaim = new RejectAllOfClaim();
+    }
+
+    if (claim.statementOfMeans) {
+      claim.statementOfMeans = new StatementOfMeans();
+    }
+
+    //clear common fields
```
TatsyMaslova, line 25: after we selected a payment option we should not clear the whole thing. we should clear the previously associated fields if they do not apply. also don't like comments in code sorry . prefer clear method names and small methods instead of big methods with comments

neeta-hmcts:
closing this pull request as it got messed up with other commits.  I have addressed the given comments for this PR and will raise a new PR for same.

PR #1066, opened 2022-11-02
Initialise Claimant Response Model and Service

commitBlob:
### JIRA link ###
https://tools.hmcts.net/jira/browse/CIV-5540

### Change description ###
Part of CIV-5540

This is not complete ticket, this is only initial config in order for other tickets to build on top of it

```
[ ] Yes
[x] No
```


laarmada (approved):

src/main/services/features/claimantResponse/claimantResponseService.ts
```diff
@@ -0,0 +1,48 @@
+import {ClaimantResponse} from '../../../common/models/claimantResponse';
+import {getCaseDataFromStore, saveDraftClaim} from '../../../modules/draft-store/draftStoreService';
+
+const {Logger} = require('@hmcts/nodejs-logging');
+const logger = Logger.getLogger('claimantResponseService');
+
+const getClaimantResponse = async (claimId: string): Promise<ClaimantResponse> => {
+  try {
+    const claim = await getCaseDataFromStore(claimId);
+    return (claim?.claimantResponse) ? claim.claimantResponse : new ClaimantResponse();
```
GorilaNaranja, line 10: I think we don't need ? here, getCaseDataFromStore will always return a new Claim().

`return (claim.claimantResponse) ? claim.claimantResponse : new ClaimantResponse();`

lfazelyniaHMCTS (approved):

kenneth-hmcts (approved):

GorilaNaranja (approved):
A minor comment but approved

src/main/services/features/claimantResponse/claimantResponseService.ts
```diff
@@ -0,0 +1,48 @@
...
+
+const saveClaimantResponse = async (claimId: string, value: any, claimantResponsePropertyName: string, parentPropertyName?: string): Promise<void> => {
+  try {
+    const claim: any = await getCaseDataFromStore(claimId);
+    if (claim.claimantResponse) {
+      if (parentPropertyName && claim.claimantResponse[parentPropertyName]) {
+        claim.claimantResponse[parentPropertyName][claimantResponsePropertyName] = value;
+      } else if (parentPropertyName && !claim.claimantResponse[parentPropertyName]) {
+        claim.claimantResponse[parentPropertyName] = {[claimantResponsePropertyName]: value};
+      } else {
```
laperla11, line 25: while working on my story I found out that we need to add another if condition here when there is no parent property;
`else if (!parentPropertyName) {
 claim.claimantResponse[claimantResponsePropertyName] = value;
 }
`

laperla11 (approved):

PR #1086, opened 2022-11-07
add other witheness to check you answer page

kamilewandowski:
### JIRA link ###
https://tools.hmcts.net/jira/browse/CIV-5493

### Change description ###
add other witnesses to check answer page

```
[ ] Yes
[X] No
```


laarmada (commented):
Where are the related test?

src/main/services/features/response/checkAnswers/hearingRequirementSection/buildHearingRequirementSection.ts
```diff
@@ -0,0 +1,49 @@
...
+
+const getEmptyStringIfUndefined = (value: string): string => value || "";
+
+const getWitnesses = (claim: Claim, claimId: string, lang: string | unknown): SummaryRow [] => {
+  const witnessesHref = constructResponseUrlWithIdParams(claimId, DQ_DEFENDANT_WITNESSES_URL)
+  const witnesses: OtherWitnessItems[] = claim.directionQuestionnaire.witnesses.otherWitnesses.witnessItems;
+  const summaryRows: SummaryRow [] = [];
+
+  witnesses.map((item, index, items) => {
+    const witnessDetails = t('COMMON.INPUT_LABELS.FIRST_NAME') + ": " + getEmptyStringIfUndefined(item.firstName) + "<br />" +
```
TatsyMaslova, line 20: I am not sure this is the way we usually show an array of elements in the summary page. Please have a look at accounts or court orders. usually the field names are not in the value bit but in the label section. and we have change link against each. so in this case would be Witness 1 and under it would be 'First Name' in the label section and witness first name in the value section and change link

src/main/modules/i18n/locales/en.json
```diff
@@ -318,6 +320,8 @@
       "FREE_TELEPHONE_CONTACT_NUMBER": "Contact number",
       "FREE_TELEPHONE_MEDIATION": "Will you try free mediation",
       "FIRST_PAYMENT": "Make the first payment on",
+      "HEARING_REQUIREMENTS_TITLE": "Hearing requirements",
+      "FIRST_NAME": "First name",
```
laperla11, line 324: no need to introduce new key, it is already under COMMON --> INPUT_LABELS

src/main/modules/i18n/locales/en.json
```diff
@@ -330,13 +334,15 @@
       "INCOME_JOB_SEEKER_ALLOWANCE": "Jobseeker's Allowance (income based)",
       "INCOME_JOB_SEEKER_ALLOWANCE_CONTRIBUTION": "Jobseeker's Allowance (contribution based)",
       "INCOME_WORKING_TAX_CREDIT": "Working Tax Credit",
+      "LAST_NAME": "Last name",
```
laperla11, line 337: same here

laperla11 (commented):
your unit tests are missing, also new keys in en.json should also be included to cy.json file with the englist text for now.

src/main/services/features/response/checkAnswers/hearingRequirementSection/buildHearingRequirementSection.ts
```diff
@@ -0,0 +1,46 @@
...
+const changeLabel = (lang: string | unknown): string => t('COMMON.BUTTONS.CHANGE', {lng: getLng(lang)});
+
+const getEmptyStringIfUndefined = (value: string): string => value || '';
+
+const getWitnesses = (claim: Claim, claimId: string, lang: string | unknown): SummaryRow [] => {
+  const witnessesHref = constructResponseUrlWithIdParams(claimId, DQ_DEFENDANT_WITNESSES_URL);
+  const witnesses: OtherWitnessItems[] = claim.directionQuestionnaire.witnesses.otherWitnesses.witnessItems;
+  const summaryRows: SummaryRow [] = [];
+
+  witnesses.map((item, index, items) => {
```
TatsyMaslova, line 20: I am not sure whether i already mentioned this, but looks like I didnt. I think we should stick to how we already do these arrays - if you look at examples on accounts and court orders. otherwise we have three /four different ways

src/main/modules/i18n/locales/en.json
kamilewandowski: done

src/main/modules/i18n/locales/en.json
kamilewandowski: done

src/main/modules/i18n/locales/en.json
```diff
@@ -307,8 +307,10 @@
       "DETAILS_TITLE": "Your details",
       "DETAILS_TITLE_CLAIMANT": "Your details (claimant)",
       "DISABILITY_ARE_YOU_DISABLED": "Are you disabled?",
+      "DO_YOU_HAVE_OTHER_WITNESSES": "Do you have other witnesses?",
```
GorilaNaranja, line 310: Is this text missing in cy.json?

kamilewandowski: fixed

RosDiaf (approved):

kenneth-hmcts (approved):

src/main/modules/i18n/locales/en.json
```diff
@@ -307,8 +307,10 @@
...
       "DOB": "Date of birth",
       "EMAIL": "Email",
+      "EMAIL_ADDRESS": "Email",
```
laperla11, line 313: Should be `Email address` according to the cy.son also you already have email on previous line

kamilewandowski: yes it should be ```Email address``` just fixed it

laperla11 (approved):

GorilaNaranja (approved):

PR #1113, opened 2022-11-14
Civ 5493 hearing requirement witnesses

kamilewandowski:
### JIRA link ###
https://tools.hmcts.net/jira/browse/CIV-5493

### Change description ###
update witnesses view

```
[ ] Yes
[X] No
```


src/main/services/features/response/checkAnswers/hearingRequirementSection/buildHearingRequirementSection.ts
```diff
@@ -0,0 +1,46 @@
...
+
+const getEmptyStringIfUndefined = (value: string): string => value || '';
+
+const getWitnesses = (claim: Claim, claimId: string, lang: string | unknown): SummaryRow [] => {
+  const witnessesHref = constructResponseUrlWithIdParams(claimId, DQ_DEFENDANT_WITNESSES_URL);
+  const witnesses: OtherWitnessItems[] = claim.directionQuestionnaire.witnesses.otherWitnesses.witnessItems;
+  const summaryRows: SummaryRow [] = [];
+  let index = 0;
+
+  for(const witness of witnesses) {
```
GorilaNaranja, line 21: Just a suggestion, if you use forEach you will have index for each iteration and don't need to +1 index.

```
witnesses.forEach((witness, index) =>{
  summaryRows.push(summaryRow(`${t('PAGES.CHECK_YOUR_ANSWER.WITNESS', {lng: getLng(lang)})} ${index}`, '', witnessesHref, changeLabel(lang)));
})
```

src/main/services/features/response/checkAnswers/hearingRequirementSection/buildHearingRequirementSection.ts
```diff
@@ -0,0 +1,46 @@
...
+import {Claim} from 'models/claim';
+import {t} from 'i18next';
+import {getLng} from '../../../../../common/utils/languageToggleUtils';
+import {SummaryRow, summaryRow} from '../../../../../common/models/summaryList/summaryList';
+import {constructResponseUrlWithIdParams} from '../../../../../common/utils/urlFormatter';
+import {DQ_DEFENDANT_WITNESSES_URL} from '../../../../../routes/urls';
+import {OtherWitnessItems} from '../../../../../common/models/directionsQuestionnaire/witnesses/otherWitnessItems';
+import {YesNo} from '../../../../../common/form/models/yesNo';
+
+const changeLabel = (lang: string | unknown): string => t('COMMON.BUTTONS.CHANGE', {lng: getLng(lang)});
```
GorilaNaranja, line 11: I know this is used in other cases, but is unknow needed? I think function getLng will always return a string, if it doesn't know the language it will return 'en' as default.

src/main/services/features/response/checkAnswers/hearingRequirementSection/buildHearingRequirementSection.ts
```diff
@@ -0,0 +1,46 @@
...
+
+const getEmptyStringIfUndefined = (value: string): string => value || '';
+
+const getWitnesses = (claim: Claim, claimId: string, lang: string | unknown): SummaryRow [] => {
+  const witnessesHref = constructResponseUrlWithIdParams(claimId, DQ_DEFENDANT_WITNESSES_URL);
+  const witnesses: OtherWitnessItems[] = claim.directionQuestionnaire.witnesses.otherWitnesses.witnessItems;
```
GorilaNaranja, line 17: Could happen that some of the properties doesn't exist? I'm not sure if we should add question mark to them or not.

GorilaNaranja (approved):

src/main/services/features/response/checkAnswers/hearingRequirementSection/buildHearingRequirementSection.ts
kamilewandowski: I do not think the `?` are need in here cause `getWitnesses` is called in `if` statement where I check if that is empty. So it should not cause any issues.

src/main/services/features/response/checkAnswers/hearingRequirementSection/buildHearingRequirementSection.ts
kamilewandowski: I have removed that `unknown` if `getLng` return default value

kenneth-hmcts (approved):

PR #1192, opened 2022-12-09
Civ 5614 launch darkly

KyleMulvihillV1:
### JIRA link ###

https://tools.hmcts.net/jira/browse/CIV-5614

### Change description ###

LaunchDarkly client, with temporary  feature toggle added for CUI

```
[ ] Yes
[ ] No
```


laperla11 (approved):

src/main/app/auth/launchdarkly/launchDarklyClient.ts
```diff
@@ -0,0 +1,24 @@
+import * as LaunchDarkly from 'launchdarkly-node-server-sdk';
```
TatsyMaslova, line 1: as mentioned before could you import only what's needed instead of everything

src/main/server.ts
```diff
@@ -14,12 +16,20 @@ if (app.locals.ENV === 'development') {
     cert: readFileSync(path.join(sslDirectory, 'localhost.crt')),
     key: readFileSync(path.join(sslDirectory, 'localhost.key')),
   };
-  const server = createServer(sslOptions, app);
-  server.listen(port, () => {
-    logger.info(`Application started: https://localhost:${port}`);
-  });
+  getFlagValue('cui-enabled').then(flagValue=>{
```
TatsyMaslova, line 19: it is better to use await

src/test/unit/app/auth/launchdarkly/launchDarklyClient.test.ts
```diff
@@ -0,0 +1,16 @@
+import * as LaunchDarkly from 'launchdarkly-node-server-sdk';
```
TatsyMaslova, line 1: I don't see the point in creating this test if we are not testing anything . however would have tested where you have made changes in server.ts

src/main/server.ts
TatsyMaslova: also if cui is not enabled we cannot start the application? i don't see the point in that. i thought lauch darkley is for enabling some features so that we don't mess release 1. however what do we achieve in not being able to start the server?

sampankumar: We actually dont need this code. As this ticket was only to be able to connect to LD and get the flag value. This code can be removed.

src/test/unit/app/auth/launchdarkly/launchDarklyClient.test.ts
TatsyMaslova: removed test

TatsyMaslova (approved):

PR #1217, opened 2022-12-22
CIV-5673 Claimant mediation phone number

laperla11:
### JIRA link ###
https://tools.hmcts.net/jira/browse/CIV-5673

### Change description ###
Refactor following pages to use for claimant;
- Free telephone mediation screen
- Enter a phone number/confirm your telephone number

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/services/features/response/mediation/mediationService.ts
+++ b/src/main/services/features/response/mediation/mediationService.ts
@@ -1,5 +1,6 @@
 import {getCaseDataFromStore, saveDraftClaim} from '../../../../modules/draft-store/draftStoreService';
 import {Mediation} from '../../../../common/models/mediation/mediation';
+import {ClaimantResponse} from 'common/models/claimantResponse';
 
 const {Logger} = require('@hmcts/nodejs-logging');
 const logger = Logger.getLogger('freeMediationService');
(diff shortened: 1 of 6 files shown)


GorilaNaranja (approved):

KyleMulvihillV1 (approved):

src/main/services/features/response/mediation/mediationService.ts
```diff
@@ -1,13 +1,27 @@
...
+import {ClaimantResponse} from 'common/models/claimantResponse';
 
 const {Logger} = require('@hmcts/nodejs-logging');
 const logger = Logger.getLogger('freeMediationService');
 
 const getMediation = async (claimId: string): Promise<Mediation> => {
   try {
     const claim = await getCaseDataFromStore(claimId);
+
+    if (claim.isClaimantIntentionPending() && claim.claimantResponse?.mediation) {
```
portizdev, line 12: This could be split into 2 ifs and we save a call to isClaimantIntentPending. Just an if on whether it is claimantIntentPending and inside another if that checks if the mediation data exists.

It works the same, but from my point of view it would be cleaner.

src/main/services/features/response/mediation/mediationService.ts
```diff
@@ -20,13 +34,26 @@ const getMediation = async (claimId: string): Promise<Mediation> => {
...
+        claim.claimantResponse = new ClaimantResponse();
+        const mediation = new Mediation();
+        mediation[mediationPropertyName] = value;
+        claim.claimantResponse.mediation = mediation;
+      }
+    } else if (claim.mediation) {
       claim.mediation[mediationPropertyName] = value;
-    } else {
-      const mediation: any = new Mediation();
+    } else if (!claim.mediation) {
```
portizdev, line 55: Is this condition necessary? Isn't it enough with the else?
If it is not ClaimantIntentPending, and it does not have claim.mediation, the else would be that there is no claim.mediation.

Correct me if I am wrong.

src/main/services/features/response/mediation/mediationService.ts
laperla11: I don't like the nested if statement so I prefer to use it in this way and for me it's more clear.

src/main/services/features/response/mediation/mediationService.ts
laperla11: That's true I could only use `else` here. done

portizdev (approved):

PR #1257, opened 2023-01-10
Rapayment length to update when typing new value

RosDiaf:
### JIRA link ###
https://tools.hmcts.net/jira/browse/CIV-6695

### Change description ###

**Bugs:**

1. After entering repayment plan and re-visit the same page, '**Length of repayment plan**' label is empty. **(Fixed)**

2. Length of repayment plan update the value when selecting radio box frequency only, not when typing different amount value for payment plan too. **(Fixed)**
 
211544034-dd868f87-9a34-4790-b665-e31913c201d6

```
[ ] Yes
[ ] No
```


RosDiaf:
**Bugs:**

1. After entering repayment plan and re-visit the same page, '**Length of repayment plan**' label is empty. **(Fixed)**

2. Length of repayment plan update the value when selecting radio box frequency only, not when typing different amount value for payment plan too. **(Fixed)**
 
bug

KyleMulvihillV1 (approved):

laperla11 (approved):

src/main/assets/js/calculate-length-repayment.js
```diff
@@ -45,12 +45,22 @@
   hideAll();
 
   if (repaymentFrequency) {
+
+    window.addEventListener('load', () => {
+      repaymentFrequencyValue = repaymentFrequency.querySelector(repaymentFrequencyRadioBox);
+      if (repaymentFrequencyValue && repaymentFrequencyValue.value) {
```
TatsyMaslova, line 51: repaymentFrequencyValue?.value probably would be better especially if you are repeating the code elsewhere

RosDiaf: Added

portizdev (approved):

src/main/assets/js/calculate-length-repayment.js
```diff
@@ -45,12 +45,22 @@
   hideAll();
 
   if (repaymentFrequency) {
+
+    window.addEventListener('load', () => {
+      repaymentFrequencyValue = repaymentFrequency.querySelector(repaymentFrequencyRadioBox);
+      if (repaymentFrequencyValue && repaymentFrequencyValue?.value) {
```
TatsyMaslova, line 51: still not happy sorry. you don't need the first part. the first part is checked by the "?" so ` repaymentFrequencyValue?.value` is the same as` repaymentFrequencyValue && repaymentFrequencyValue.value` so you don't need the first part of the && if you are using ?. hope it makes sense

RosDiaf: changes added

TatsyMaslova (approved):

PR #1283, opened 2023-01-19
CIV-6063 - part admit pay instalments how they want to pay

RosDiaf:
### JIRA link ###
https://tools.hmcts.net/jira/browse/CIV-6063

### Change description ###
The scope of this ticket is to develop the "How they want to pay" section in the scenario where the defendant suggested the repayment plan and their financial details

```
[ ] Yes
[ ] No
```


src/main/common/utils/repaymentUtils.ts
```diff
@@ -69,3 +69,21 @@ export const convertFrequencyToText = (frequency: string, lng: string): string =
       return t('COMMON.FREQUENCY_OF_PAYMENTS.MONTHLY', { lng });
   }
 };
+
+export const getRepaymentLength = (claim: Claim, lng: string) => {
+  const repaymentFrequency = getRepaymentFrequency(claim);
+  let repaymentLength = '';
+  switch (repaymentFrequency) {
+    case TransactionSchedule.WEEK:
+      repaymentLength = getNumberOfInstalments(claim) === 2 ? t('COMMON.SCHEDULE.TWO_WEEKS', { lng }) : getNumberOfInstalments(claim) + ' ' + t('COMMON.SCHEDULE.WEEKS_LOWER_CASE', { lng });
```
laperla11, line 78: Do we need to check the state repaymentPlan  is `one week` only?

laperla11 (dismissed):

src/main/common/utils/repaymentUtils.ts
RosDiaf: No we don't. Because based on the model validation the payment amount has to be less of the amount they own if part admit or total amount if full admit. So the number of instalments will always be minimum 2

src/main/common/utils/repaymentUtils.ts
```diff
@@ -69,3 +69,21 @@ export const convertFrequencyToText = (frequency: string, lng: string): string =
       return t('COMMON.FREQUENCY_OF_PAYMENTS.MONTHLY', { lng });
   }
 };
+
+export const getRepaymentLength = (claim: Claim, lng: string) => {
+  const repaymentFrequency = getRepaymentFrequency(claim);
```
GorilaNaranja, line 74: There is a function in Luxon to calculate duration between 2 dates, maybe it's a easiest and solid way to do it.

https://stackoverflow.com/questions/63763221/how-to-calculate-a-duration-between-two-dates-in-luxon

src/main/services/features/claimantResponse/claimantResponseService.ts
```diff
@@ -55,6 +58,62 @@ const saveClaimantResponse = async (claimId: string, value: any, claimantRespons
   }
 };
 
+const constructRepaymentPlanSection = (claim: Claim, lng: string) => {
+  const sectionRows = [];
+
+  if (claim.partialAdmission?.alreadyPaid?.option === YesNo.NO && claim.partialAdmission?.paymentIntention?.paymentOption === PaymentOptionType.INSTALMENTS) {
```
GorilaNaranja, line 64: We have some claim methods to get that kind of info easier, for example: claim.isPAPaymentOptionInstallments()

src/main/services/features/claimantResponse/claimantResponseService.ts
```diff
@@ -419,6 +478,20 @@ const constructCourtOrdersSection = (claim: Claim, lng: string) => {
   return sectionRows;
 };
 
+const hasDebtAmount = (priorityDebts: PriorityDebts) => {
```
GorilaNaranja, line 481: return type

src/main/services/features/claimantResponse/claimantResponseService.ts
```diff
@@ -55,6 +58,62 @@ const saveClaimantResponse = async (claimId: string, value: any, claimantRespons
   }
 };
 
+const constructRepaymentPlanSection = (claim: Claim, lng: string) => {
```
GorilaNaranja, line 61: return type

src/main/services/features/claimantResponse/claimantResponseService.ts
```diff
@@ -419,6 +478,20 @@ const constructCourtOrdersSection = (claim: Claim, lng: string) => {
...
 };
 
+const hasDebtAmount = (priorityDebts: PriorityDebts) => {
+  let amountFlag = false;
+  if (priorityDebts) {
+    const priorityDebtsKeys = Object.keys(priorityDebts);
+    priorityDebtsKeys.forEach((priorityDebtKey) => {
+      const priorityDebt = priorityDebts[priorityDebtKey as keyof PriorityDebts];
+      if (priorityDebt?.transactionSource?.amount) {
+        amountFlag = true;
```
GorilaNaranja, line 488: Instead wait for all loops you can return directly true and finish the loop beacuse if at least one of them is true you will return true. 

Also, I don't understand this function very well. I think for your purpose you can use "modern" JS methods like .find or .filter to find if one object has that property within the array.

src/main/common/utils/repaymentUtils.ts
```diff
@@ -69,3 +69,21 @@ export const convertFrequencyToText = (frequency: string, lng: string): string =
       return t('COMMON.FREQUENCY_OF_PAYMENTS.MONTHLY', { lng });
   }
 };
+
+export const getRepaymentLength = (claim: Claim, lng: string) => {
```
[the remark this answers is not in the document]
RosDiaf: Added

src/main/services/features/claimantResponse/claimantResponseService.ts
RosDiaf: isPAPaymentOptionInstallments added

src/main/services/features/claimantResponse/claimantResponseService.ts
RosDiaf: Added

src/main/services/features/claimantResponse/claimantResponseService.ts
RosDiaf: Added

src/main/services/features/claimantResponse/claimantResponseService.ts
RosDiaf: This return statement does not exit the forEach
I have changed to for loop where I can exit with break keyword
js method like .find or .filter works on arrays. priorityDebts in Redis is saved as an object

GorilaNaranja (approved):

GorilaNaranja (approved):

laarmada (approved):

PR #1377, opened 2023-02-20
CIV-7187 bug adding row

GorilaNaranja:
### JIRA link ###
https://tools.hmcts.net/jira/browse/CIV-7187

### Change description ###
Bug adding another element (append-row)

```
[x] Yes
[ ] No
```


GorilaNaranja:
This changes affect to all pages that use "Add another ..." and "Remove" button. For example:

http://localhost:3001/case/1645882162449409/directions-questionnaire/availability-dates
http://localhost:3001/case/1645882162449409/directions-questionnaire/support-required
http://localhost:3001/case/1645882162449409/response/statement-of-means/bank-accounts
http://localhost:3001/case/1653466401017331/directions-questionnaire/expert-details

laarmada (approved):

src/main/assets/js/append-row.js
```diff
@@ -104,21 +107,42 @@ document.addEventListener('DOMContentLoaded', function () {
...
+          addEventToUnavailableDatesRadios(element);
+        }
       });
     }
   }
 
+  function getNumberFromElementName() {
+    const elements = document.getElementsByClassName('multiple-row');
+    const lastItem = elements[elements.length-1];
+    const lastRadioInput = lastItem.getElementsByClassName('govuk-radios__input');
```
PaulH-HMCTS, line 131: Seeing as all we are doing is extracting the id, it would be more flexible to match `lastItem.getElementsByTagName('input')` (and maybe also select) - did you try this? If it works we won't need a special case for every input type

GorilaNaranja: My concern was about the select but it is working because the page that use selects also has a input so you were right. It seems work getting the element by tag input. I'm pushing the changes.

LeonardoPalmeiro (approved):

LeonardoPalmeiro (approved):

PaulH-HMCTS (approved):

src/main/assets/js/append-row.js
GorilaNaranja: @PaulH-HMCTS  we found that functional tests fail if get the number from "inputs".
I have to revert these change and keep this very specific fix only for radios and checkbox (instead get generic inputs). If I use the generic approach functional test are failing because for some pages ids are like id="test-1-amount" but for others are id="test[1][amount]" or id="test[field][1][amount]"... so I'm going to keep the specific fix for 2 scenarios that were failin, radios and checkboxes.

laperla11 (approved):

LeonardoPalmeiro (approved):

portizdev (approved):

PR #1410, opened 2023-02-28
CIV-0000 Addition of a test that ensures consistency of keys between en and cy language files

PaulH-HMCTS:
### JIRA link ###
N/A

### Change description ###
A new test has been added that reads the translation keys from the en.json and cy.json language translation files. Then a simple check is carried out to ensure that the two files have the same keys, in the same structure and order.
I have fixed a few issues in both the language files, mostly due to differences in the ordering but there were some actual problems with key names or structure that were also uncovered.

```
[ ] Yes
[x] No
```


LeonardoPalmeiro (approved):

portizdev (approved):
Very good solution to a common problem of the project!

TatsyMaslova (approved):

PR #1478, opened 2023-03-28
CIV-7603 add warnings to a11y test

laarmada:
### JIRA link ###

https://tools.hmcts.net/jira/browse/CIV-7603

### Change description ###

add coverage for warning in accessibility tests

```
[ ] Yes
[x] No
```


src/test/a11y/a11y.test.ts
```diff
@@ -65,13 +67,232 @@ function expectNoErrors(messages: PallyIssue[]): void {
...
+  const nonTextContent  = messages.filter(m => m.code.includes('WCAG2AA.Principle1.Guideline1_1'));
+  if (nonTextContent.length > 0) {
+    const errorsAsJson = `${JSON.stringify(nonTextContent, null, 2)}`;
+    fail(`There are non text content Guideline 1.1.1 issues: \n${errorsAsJson}\n`);
+  }
+
+}
+
+function expectNoAdaptableWarnings(messages: PallyIssue[]): void {
+  //const info  = messages.filter(m => m.code.includes('WCAG2AA.Principle1.Guideline1_3.1_3_1'));
```
laperla11, line 80: Do you need the commented codes?

src/test/a11y/a11y.test.ts
```diff
@@ -80,6 +301,13 @@ function testAccessibilityWithActions(url: string, actions: string[]): void {
 }
 
 function testAccessibility(url: string): void {
+  //we need to add actions for example:
```
laperla11, line 304: Should we do it part of another task?

laperla11 (changes requested):

src/test/a11y/a11y.test.ts
laarmada: exactly, only looking to make easy the next issue.

src/test/a11y/a11y.test.ts
laarmada: if we uncomment that lines we have accessibility issue so I will create an specific issue to fix each commented section.

PR #1496, opened 2023-03-31
CIV-7500 pcq journey on check and submit

kamilewandowski:
### JIRA link ###
https://tools.hmcts.net/jira/browse/CIV-7500

### Change description ###
add protected characteristics questions

```
[ ] Yes
[X] No
```


neeta-hmcts (approved):
Looks Fine to me.

laperla11 (dismissed):

src/main/routes/features/response/checkAnswersController.ts
```diff
@@ -34,7 +35,12 @@ checkAnswersController.get(RESPONSE_CHECK_ANSWERS_URL,
     try {
       const claim = await getCaseDataFromStore(req.params.id);
       const form = new GenericForm(getStatementOfTruth(claim));
-      renderView(req, res, form, claim);
+
+      if(!claim.pcqPageSeen) {
+        res.redirect(constructResponseUrlWithIdParams(req.params.id, PCQ_URL));
```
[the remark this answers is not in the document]
TatsyMaslova: I think it needs to be covered by a different story. we need to ask Simon to create one

src/main/views/features/response/pcq.njk
```diff
@@ -0,0 +1,46 @@
...
+      <p class="govuk-body">
+        {{ t('PAGES.PCQ.YOUR_ANSWER_WILL_HELP') }}
+      </p>
+      <p class="govuk-body">
+        {{ t('PAGES.PCQ.FIND_OUT_HOW_WE_USE') }}
+        {{ externalLink(t, 'PAGES.PCQ.PRIVACY_POLICY', '', false) }}
+      </p>
+      <div class="form-group">
+        {{ govukButton({
+          text: t('PAGES.PCQ.CONTINUE_TO_QUESTIONS'),
```
TatsyMaslova, line 35: this should be handled in controller. it should redirect to next url. even if the pcq questions do not exist we still do this and include in test. this would be handled in the post in the controller as well as setting the pcq to true should be in the post and saved in redis

src/main/views/features/response/pcq.njk
```diff
@@ -0,0 +1,46 @@
...
+          href: ''
+        }) }}
+        {{ govukButton({
```
TatsyMaslova, line 38: same applies here. if they dont want to answer this should be handled in the controller with post and the pcq flag should be false

src/main/views/features/response/pcq.njk
TatsyMaslova:
```
GIVEN As a Individual or Sole trader

AND defendant completes all the required steps of their response

WHEN click on "check and submit your response" for the first time

THEN open PCQ journey page
```
don't think this is handled... but should be according to the ACs

PR #1533, opened 2023-04-18
CIV-7982 Initial yarn 3 migration changes

PaulH-HMCTS:
### JIRA link ###
[CIV-7982](https://tools.hmcts.net/jira/browse/CIV-7982)

### Change description ###
Upgrade to yarn v3

```
[ ] Yes
[x] No
```


Dockerfile
```diff
@@ -1,7 +1,7 @@
 # ---- Base image ----
 FROM hmctspublic.azurecr.io/base/node:16-alpine as base
 COPY --chown=hmcts:hmcts . .
-RUN yarn install --production \
```
portizdev, line 4: Does this change anything from the way we used to do yarn install in the local environment? Or is this just a necessary change for docker.

LeonardoPalmeiro (approved):

Dockerfile
PaulH-HMCTS: No, we can still use `yarn install` locally, it is only the production install that is affected. The --production flag has been deprecated in yarn 3 so had to remove it and replace with `yarn workspaces focus --all --production`. This should be equivalent so shouldn't cause any problems hopefully.

TatsyMaslova (approved):

hmcts-version1-ignacio (approved):

PR #1806, opened 2023-06-29
CIV-7549 first contact links

TatsyMaslova:
### JIRA link ###

https://tools.hmcts.net/jira/browse/CIV-7549

### Change description ###

add ocmc link to privacy policy (in preview this will point to aat ocmc)
ensure we are not redirecting the user to IDAM when trying to download the document pdf

```
[ ] Yes
[ ] No
```


kenneth-hmcts (approved):

laperla11 (approved):

src/main/routes/features/document/documentDownloadController.ts
```diff
@@ -7,15 +7,16 @@ import {convertToDocumentType} from 'common/utils/documentTypeConverter';
...
 
 const documentDownloadController = Router();
 const civilServiceApiBaseUrl = config.get<string>('services.civilService.url');
-const civilServiceClient: CivilServiceClient = new CivilServiceClient(civilServiceApiBaseUrl);
+
 const civilServiceClientForDocRetrieve: CivilServiceClient = new CivilServiceClient(civilServiceApiBaseUrl, true);
 
 documentDownloadController.get(CASE_DOCUMENT_DOWNLOAD_URL, async (req: Request, res: Response, next: NextFunction) => {
   try {
-    const claim: Claim = await civilServiceClient.retrieveClaimDetails(req.params.id, <AppRequest>req);
```
neeta-hmcts, line 18: we are using same controller for  downloading SDO order.. plz make sure this change will not impact that.

TatsyMaslova: The new change that was added for SDO broke existing functionality to be able to download the pdf without being logged in (i.e. first contact journey) I will ask to regression test this with SDO too. we need both functionalities working. An alternative would be to update the case when the sdo is drawn before we call the download document. in the end we need to diplay the link only when the link exists in the case

PR #1812, opened 2023-06-30
CIV-9460 footer links

GorilaNaranja:
### JIRA link ###
https://tools.hmcts.net/jira/browse/CIV-9460

### Change description ###
CUI R1 - Footer Links update

```
[ ] Yes
[x] No
```


src/main/modules/nunjucks/index.ts
```diff
@@ -143,6 +143,10 @@ export class Nunjucks {
     nunjucksEnv.addGlobal('today', new Date());
     nunjucksEnv.addGlobal('nextMonth', nextMonth);
     nunjucksEnv.addGlobal('PaymentOptionType', PaymentOptionType);
+    nunjucksEnv.addGlobal('ContactUsUrl', 'https://www.moneyclaims.service.gov.uk/contact-us');
```
laperla11, line 146: You can also add them to `external-urls.njk` file and import form there instead of creating global variables as they're not widely used.

laperla11 (approved):

src/main/modules/nunjucks/index.ts
TatsyMaslova: sorry Felipe but link to ocmc should be dynamic. please have a look at my recent pr that is in deploy awaits. also have a look how we are doing it in dashboard. the reason is that in live the url to ocmc might change. and cui will become moneyclaims. we want to configure the change just once and it should work for all links that are in cui

TatsyMaslova: https://github.com/hmcts/civil-citizen-ui/pull/1806

config/custom-environment-variables.yaml
```diff
@@ -25,6 +25,8 @@ services:
     serviceAuthTokenGeneratorUrl: SERVICE_AUTH_TOKEN_GENERATOR_URL
   cmc:
     url: OCMC_URL
+  moneyClaims:
```
TatsyMaslova, line 28: we are just using cmc. no need to create new moneyclaims

src/main/modules/nunjucks/index.ts
GorilaNaranja: Tanya I updated the code with your changes

GorilaNaranja: Kadir, I adde urls there because:
-we are using that urls in 2 different views
-as we get urls from config, I need a ts to get that info. Also, these 2 views doesn't has a controller to get the config url and pass the variables to the view, so I had to put them in a global place.

config/custom-environment-variables.yaml
GorilaNaranja: done

laperla11 (approved):

TatsyMaslova (approved):

PR #1953, opened 2023-07-25
CIV-9534 create new claim link

TatsyMaslova:
### JIRA link ###

https://tools.hmcts.net/jira/browse/CIV-9534

### Change description ###

```
[ ] Yes
[ ] No
```


src/main/common/models/dashboard/dashboardItem.ts
```diff
@@ -157,3 +158,18 @@ export const translate = (translationKey: string, params?: DashboardStatusTransl
   return t(translationKey, {lng:getLng(lang)} );
 };
 
+export const toDraftClaimDashboardItem = (claim: Claim): DashboardClaimantItem | undefined  =>{
+  if(claim == undefined || claim?.isEmpty()) {
```
GorilaNaranja, line 162: Use triple ===, but I will use !claim because it will check also if claim is null.

src/main/routes/features/dashboard/dashboardController.ts
```diff
@@ -3,19 +3,19 @@ import {Response, Router} from 'express';
...
 const civilServiceApiBaseUrl = config.get<string>('services.civilService.url');
 const civilServiceClient: CivilServiceClient = new CivilServiceClient(civilServiceApiBaseUrl);
-const ocmcBaseUrl = config.get<string>('services.cmc.url');
 
-function renderPage(res: Response, claimsAsClaimant: DashboardClaimantItem[], claimDraftSaved: DashboardClaimantItem,
+async function renderPage(res: Response, claimsAsClaimant: DashboardClaimantItem[], claimDraftSaved: DashboardClaimantItem,
   claimsAsDefendant: DashboardDefendantItem[], responseDraftSaved: boolean,
-  paginationArgumentClaimant: object, paginationArgumentDefendant: object, lang: string | unknown): void {
+  paginationArgumentClaimant: object, paginationArgumentDefendant: object, lang: string | unknown): Promise<void> {
+  const draftClaimUrl = await createDraftClaimUrl();
```
GorilaNaranja, line 18: can you please move this to the get controller? Function renderPage should only receive data and pass it to the view. I think like that you don't need to convert renderPage to an async function.

src/main/routes/features/dashboard/dashboardController.ts
```diff
@@ -36,12 +36,12 @@ dashboardController.get(DASHBOARD_URL, async function (req, res, next) {
...
   try{
     const claimsAsClaimant : DashboardClaimantItem[] = await civilServiceClient.getClaimsForClaimant(appRequest);
+    const claimDraftSaved = await getDraftClaim(user?.accessToken);
     const claimsAsDefendant: DashboardDefendantResponse = await civilServiceClient.getClaimsForDefendant(appRequest);
     const claimsAsDefendantPaginationList = buildPagination(claimsAsDefendant.totalPages, req.query?.page as string, lang);
-    const claimDraftSaved = await getOcmcDraftClaims(user?.accessToken);
     const responseDraftSaved = false;
     const paginationArgumentClaimant: object = {};
-    renderPage(res, claimsAsClaimant, claimDraftSaved, claimsAsDefendant.claims, responseDraftSaved, paginationArgumentClaimant, claimsAsDefendantPaginationList, lang);
+    await renderPage(res, claimsAsClaimant, claimDraftSaved, claimsAsDefendant.claims, responseDraftSaved, paginationArgumentClaimant, claimsAsDefendantPaginationList, lang);
```
GorilaNaranja, line 44: And here you don't have to await for it

src/main/services/dashboard/draftClaimService.ts
```diff
@@ -0,0 +1,28 @@
...
+  const isReleaseTwoEnabled = await isCUIReleaseTwoEnabled();
+  if(isReleaseTwoEnabled) {
+    return eligibility;
+  }
+  return `${ocmcBaseUrl}/eligibility`;
+};
+
+export const getDraftClaim = async (userToken: string): Promise<DashboardClaimantItem> => {
+  const isReleaseTwoEnabled = await isCUIReleaseTwoEnabled();
+  console.log(isReleaseTwoEnabled);
```
GorilaNaranja, line 20: delete log

TatsyMaslova: oops yes sorry was debugging and forgot. good spot!

LeonardoPalmeiro (dismissed):

GorilaNaranja (approved):

LeonardoPalmeiro (dismissed):

src/main/services/dashboard/draftClaimService.ts
```diff
@@ -0,0 +1,27 @@
...
+import config from 'config';
+import {DashboardClaimantItem, toDraftClaimDashboardItem} from 'models/dashboard/dashboardItem';
+import {getOcmcDraftClaims} from 'client/legacyDraftStoreClient';
+import {getCaseDataFromStore} from 'modules/draft-store/draftStoreService';
+
+const ocmcBaseUrl = config.get<string>('services.cmc.url');
+
+export const createDraftClaimUrl = async ():Promise<string> => {
+  const eligibility = '/eligibility';
+  const isReleaseTwoEnabled = await isCUIReleaseTwoEnabled();
```
laperla11, line 11: I would prefer to call this on controller and pass it to `createDraftClaimUrl` and `getDraftClaim` methods so you don't need to call it twice.

TatsyMaslova: sorry kadir, i prefer the business logic and what is enabled or not to be in the service layer and within the draft claim service. controller only be responsible for view and model

TatsyMaslova: for example draft claim service should decide what ulr needs to be built or where to go depending on the toggle. whereas controller does not need to know about the toggle at all. just needs to display the model it receives from the service

TatsyMaslova: changed it Kadir, so that it only calls the check whether the release2 is enabled once. 😄

laperla11 (approved):

GorilaNaranja (approved):

PR #1988, opened 2023-08-03
CIV-8202 - CIV-9831 - Claim task-list guard and eligibility cookie

portizdev:
### JIRA link ###
https://tools.hmcts.net/jira/browse/CIV-9831
https://tools.hmcts.net/jira/browse/CIV-8202

### Change description ###
Creation of a cookie to be set when entering eligibility/eligible and to expire in 1h.
As long as the cookie exists, the user can access claim/task-list. 
If after 1h the user has not started to create the claim (no action has been taken in claim/task-list) they will have to go through the whole eligibility process again. If they try to access claim/task-list they will be redirected to /eligibility.

Also included is that if the user is not logged in when trying to access claim/task-list, they will be redirected after the IDAM to claim/task-list instead of /dashboard.

Use code from #1889 to be able to use the task-list.

```
[ ] Yes
[X] No
```


src/main/modules/oidc/index.ts
```diff
@@ -90,6 +98,9 @@ export class OidcMiddleware {
       if (requestIsForAssigningClaimForDefendant(req) ) {
         app.locals.assignClaimId = appReq.session.assignClaimId = <string>req.query.id;
       }
+      if (requestIsForClaimIssueTaskList(req) ) {
+        app.locals.claimIssueTasklist = true;
```
laperla11, line 102: `app.locals` is not shared with instances so this'll work on local, preview but might fail on demo and master instead either assign this value to `req.sesion` or use approach on line 99:
`app.locals.assignClaimId = appReq.session.assignClaimId = <string>req.query.id;`

laperla11 (changes requested):
Should we also need to implement R2 flag?

laperla11 (approved):

laperla11 (approved):

LeonardoPalmeiro (dismissed):

GorilaNaranja (approved):

GorilaNaranja (approved):

LeonardoPalmeiro (approved):

PR #2030, opened 2023-08-10
CIV-9921 add redis expire time

kamilewandowski:
### Change description ###
add expire time for drafts stored in redis

```
[ ] Yes
[X] No
```


TatsyMaslova:
The story says it should save the draft claim as soon as the claim is navigated from eligibility page according to the ACs of your story. does it do that?

src/main/modules/draft-store/draftStoreService.ts
```diff
@@ -6,6 +6,7 @@ import {isUndefined} from 'lodash';
 const {Logger} = require('@hmcts/nodejs-logging');
 const logger = Logger.getLogger('draftStoreService');
 
+const DRAFT_EXPIRE_TIME = 24 * 60 * 60 * 180;
```
TatsyMaslova, line 9: why do you need to do multiplication instead of just writing a result there? also perhaps we could put this value in the config so that we could change it in the config instead of going to the code

kamilewandowski: 15552000 seems to be random, and it is not clear exactly how long expire time is, and also if someone needs to change that in the future will have to calculate that again. With multiplication in place, you need to update 180 to whatever will be required.

src/main/routes/features/claim/claimTaskListController.ts
```diff
@@ -14,6 +15,12 @@ claimTaskListController.get(CLAIMANT_TASK_LIST_URL, (req: AppRequest, res: Respo
   const lang = req.query.lang ? req.query.lang : req.cookies.lang;
   getCaseDataFromStore(userId)
     .then((caseData) => {
+      if(caseData.id == undefined) {
```
TatsyMaslova, line 18: here i can see several bugs: 
1. a draft claim will never have an id so every time the claimant is navigated to task list, we override existing claim with a new claim.
2. we only want to save the claim if it does not exist in draft store (i.e. there is no claim against the user) and when the user navigated from eligibility page (i.e. we need to check for the cookies/session from eligibility page)

config/default.yaml
```diff
@@ -26,6 +26,7 @@ services:
       port: 6379
       key:
       tls: false
+      expireInDays: 180
```
TatsyMaslova, line 29: default.yaml is for our local environment. so it will work on our local but not for Raja in test or in live. You need to introduce a variable and add it to custom-evnironment-variables.yaml .

TatsyMaslova: Then add it to values.yaml

src/main/routes/features/claim/claimTaskListController.ts
```diff
@@ -4,26 +4,39 @@ import {AppRequest} from 'models/AppRequest';
...
+    try {
+      const userId = req.session?.user?.id;
+      const lang = req.query.lang ? req.query.lang : req.cookies.lang;
+      const isReleasedTwoEnabled: boolean = await isCUIReleaseTwoEnabled();
+      const caseData= await getCaseDataFromStore(userId);
+
+      if(isReleasedTwoEnabled) {
+        saveDraftClaim(null, undefined);
+      } else {
+        const claim = new Claim();
```
TatsyMaslova, line 26: no this is not right. we dont have to check release 2 here. we just need to check whether casedata is undefined or not and whether the eligiblity check has been done. what we need to do is check whether release2 is enabled in getCaseDataFromStore function and if it is enabled and there is no claim for the user id then return undefined

src/main/modules/draft-store/draftStoreService.ts
```diff
@@ -32,6 +36,11 @@ const convertRedisDataToCivilClaimResponse = (data: string) => {
  * @param claimId
  */
 export const getCaseDataFromStore = async (claimId: string): Promise<Claim> => {
+  const isReleasedTwoEnabled: boolean = await isCUIReleaseTwoEnabled();
+  if(isReleasedTwoEnabled) {
```
TatsyMaslova, line 40: no it should not just return undefined all the time. it should check whether the case is there or not and if it is not there then return undefined

src/main/modules/draft-store/draftStoreService.ts
```diff
@@ -32,6 +36,11 @@ const convertRedisDataToCivilClaimResponse = (data: string) => {
...
+    return undefined;
+  }
+
   const civilClaimResponse = await getDraftClaimFromStore(claimId);
```
TatsyMaslova, line 44: so your check should be here after you called getDraftClaimFromStore. and instead of callind const claim = new Claim - check if civilClaimResponse is undefined or civilClaimRespnse.case_data is undefined  and release2 enabled then return undefined. otherwise carry on as normal

src/main/routes/features/claim/claimTaskListController.ts
```diff
@@ -4,26 +4,36 @@ import {AppRequest} from 'models/AppRequest';
...
-    .then((caseData) => {
+claimTaskListController.get(CLAIMANT_TASK_LIST_URL, claimIssueTaskListGuard, (req: AppRequest, res: Response, next: NextFunction): void => {
+  (async () => {
+    try {
+      const userId = req.session?.user?.id;
+      const lang = req.query.lang ? req.query.lang : req.cookies.lang;
+      const caseData: Claim = await getCaseDataFromStore(userId);
+
+      if(caseData == undefined) {
+        saveDraftClaim(null, undefined);
```
TatsyMaslova, line 22: this is incorrect as well if caseData is undefined then save the caseData to draft store under the user id . no need to set dates etc. it should be done in the logic where we save the case (outside the controller). if it is defined then we dont do anything. otherwise we will override existing data

src/main/modules/draft-store/draftStoreService.ts
```diff
@@ -49,17 +59,21 @@ export const getCaseDataFromStore = async (claimId: string): Promise<Claim> => {
  */
 export const saveDraftClaim = async (claimId: string, claim: Claim) => {
   let storedClaimResponse = await getDraftClaimFromStore(claimId);
+  const draftStoreClient = app.locals.draftStoreClient;
+
   if (isUndefined(storedClaimResponse.case_data)) {
     storedClaimResponse = createNewCivilClaimResponse(claimId);
+    draftStoreClient.expire(claimId, addDaysToDate(claim.createAt, DRAFT_EXPIRE_TIME_IN_DAYS).getTime());
```
TatsyMaslova, line 66: as discussed previously you need to set createAt date if the claim is a new claim. this should happen in the service for redis because this field is redis related - defines the expiration of the claim

TatsyMaslova: you are setting it in the new draft claim response but checking on the claim you are saving

src/main/modules/draft-store/draftStoreService.ts
```diff
@@ -49,17 +59,21 @@ export const getCaseDataFromStore = async (claimId: string): Promise<Claim> => {
...
+
   storedClaimResponse.case_data = claim;
-  const draftStoreClient = app.locals.draftStoreClient;
   draftStoreClient.set(claimId, JSON.stringify(storedClaimResponse));
 };
 
 const createNewCivilClaimResponse = (claimId: string) => {
   const storedClaimResponse = new CivilClaimResponse();
   storedClaimResponse.id = claimId;
+  storedClaimResponse.case_data.createAt = new Date();
```
TatsyMaslova, line 76: would not this throw a null pointer or equivalent exception?

src/main/routes/features/claim/claimTaskListController.ts
```diff
@@ -4,26 +4,33 @@ import {AppRequest} from 'models/AppRequest';
...
-    .then((caseData) => {
+claimTaskListController.get(CLAIMANT_TASK_LIST_URL, claimIssueTaskListGuard, (req: AppRequest, res: Response, next: NextFunction): void => {
+  (async () => {
+    try {
+      const userId = req.session?.user?.id;
+      const lang = req.query.lang ? req.query.lang : req.cookies.lang;
+      const caseData: Claim = await getCaseDataFromStore(userId);
+
+      if(caseData == undefined) {
+        saveDraftClaim(userId, caseData);
```
TatsyMaslova, line 22: aren't we saving undefined here?

TatsyMaslova: perhaps if caseData is undefined we can create a new CaseData then save it under userid.

TatsyMaslova: perhaps should be this:
` if(caseData == undefined) {
        saveDraftClaim(userId, new CaseData());
}`

TatsyMaslova (dismissed):

laperla11 (approved):

src/main/routes/features/claim/claimTaskListController.ts
```diff
@@ -4,26 +4,33 @@ import {AppRequest} from 'models/AppRequest';
...
-    .then((caseData) => {
+claimTaskListController.get(CLAIMANT_TASK_LIST_URL, claimIssueTaskListGuard, (req: AppRequest, res: Response, next: NextFunction): void => {
+  (async () => {
+    try {
+      const userId = req.session?.user?.id;
+      const lang = req.query.lang ? req.query.lang : req.cookies.lang;
+      const caseData: Claim = await getCaseDataFromStore(userId);
+
+      if(caseData === undefined || caseData.id === undefined) {
+        saveDraftClaim(userId, new Claim());
```
LeonardoPalmeiro, line 22: maybe we're missing await in here

LeonardoPalmeiro (changes requested):

src/main/routes/features/claim/claimTaskListController.ts
```diff
@@ -4,26 +4,33 @@ import {AppRequest} from 'models/AppRequest';
...
-  getCaseDataFromStore(userId)
-    .then((caseData) => {
+claimTaskListController.get(CLAIMANT_TASK_LIST_URL, claimIssueTaskListGuard, (req: AppRequest, res: Response, next: NextFunction): void => {
+  (async () => {
+    try {
+      const userId = req.session?.user?.id;
+      const lang = req.query.lang ? req.query.lang : req.cookies.lang;
+      const caseData: Claim = await getCaseDataFromStore(userId);
+
+      if(caseData === undefined || caseData.id === undefined) {
```
laperla11, line 21: `caseData.?id === undefined` is covering both conditions

laperla11 (dismissed):

kenneth-hmcts (dismissed):

src/main/routes/features/claim/claimTaskListController.ts
```diff
@@ -1,29 +1,36 @@
...
-      next(error);
-    });
-});
+claimTaskListController.get(CLAIMANT_TASK_LIST_URL, claimIssueTaskListGuard, (async (req: AppRequest, res: Response, next: NextFunction) => {
+  try {
+    const userId = req.session?.user?.id;
+    const lang = req.query.lang ? req.query.lang : req.cookies.lang;
+    const draftClaim = await getDraftClaimFromStore(userId);
+    if (!draftClaim?.case_data) {
+      await creteDraftClaimInStoreWithExpiryTime(userId);
```
TatsyMaslova, line 23: spelling mistake on the function - create not crete 😄 . i also do lots of spelling mistakes

src/main/modules/draft-store/draftStoreService.ts
```diff
@@ -66,3 +73,11 @@ const createNewCivilClaimResponse = (claimId: string) => {
 export const deleteDraftClaimFromStore = async (claimId: string): Promise<void> => {
   await app.locals.draftStoreClient.del(claimId);
 };
+
+export async function creteDraftClaimInStoreWithExpiryTime(claimId: string) {
```
TatsyMaslova, line 77: spelling mistake on the method. should be create

LeonardoPalmeiro (dismissed):

GorilaNaranja (approved):

neeta-hmcts (approved):

src/main/routes/features/claim/claimTaskListController.ts
laperla11: good spot - fixed the spelling

src/main/modules/draft-store/draftStoreService.ts
laperla11: fixed the spelling

TatsyMaslova (approved):

laperla11:
replaced ny PR-2158

PR #2248, opened 2023-10-02
CIV-10046 - BUG ITHC - Document Security

neeta-hmcts:
### JIRA link ###
https://tools.hmcts.net/jira/browse/CIV-10046

### Change description ###
Fixed document download security issue. 

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/app/client/civilServiceClient.ts
+++ b/src/main/app/client/civilServiceClient.ts
@@ -201,10 +201,11 @@ export class CivilServiceClient {
     }
   }
 
-  async retrieveDocument(documentId: string) {
+  async retrieveDocument(req: AppRequest, documentId: string ) {
+    const config = this.getConfig(req);
     try {
       const response: AxiosResponse<object> = await this.client.get(CIVIL_SERVICE_DOWNLOAD_DOCUMENT_URL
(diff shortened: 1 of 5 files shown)


src/main/app/client/civilServiceClient.ts
```diff
@@ -201,10 +201,11 @@ export class CivilServiceClient {
     }
   }
 
-  async retrieveDocument(documentId: string) {
+  async retrieveDocument(req: AppRequest, documentId: string ) {
```
laperla11, line 204: You don't need `documentId` here as it's embedded in `Request` --> `req.params.documentId`

laperla11 (dismissed):

TatsyMaslova:
This change then makes it impossible for the first journey to download claim form. is this what we want? and if so do we need to remove the link to pdf in the first journey (where the defendant is not logged in)

neeta-hmcts:
Yes.. That's correct... we are removing the download claim from form first-contact claim summary page https://github.com/hmcts/civil-citizen-ui/pull/2252

TatsyMaslova (approved):

laperla11 (dismissed):

PR #2255, opened 2023-10-02
CIV-10391 Respondent1 response deadline date update

kalachandrasekar1:
JIRA link (if applicable)
https://tools.hmcts.net/jira/browse/CIV-10391

Change description
respondent1 response deadline date updated

```
[ ] Yes
[ ] No
```


src/main/routes/features/claim/claimFeeController.ts
```diff
@@ -14,9 +14,7 @@ claimFeeController.get(CLAIM_FEE_URL, (async (req, res, next: NextFunction) => {
     const claim = await getClaimById(claimId, req);
     claim.issueDate = new Date();
     // Need change to use the real deadline
-    const date = new Date();
-    date.setDate(date.getDate() + 28);
-    claim.respondent1ResponseDeadline = date;
+    claim.respondent1ResponseDeadline = claim.issueDate;
```
TatsyMaslova, line 17: I do not agree that we need to set the deadline in cui. I think it needs to be done in the backend when the claim is actually submitted after payment. the reason is that when you set the deadline date but something goes wrong in the backend during the submission of the claim after payment - then the deadline for the defendant is set but the defendant does not get the claim. this is similar to issue date. but issue date is not part of your story.

Jenkinsfile_CNP
```diff
@@ -7,10 +7,10 @@ import uk.gov.hmcts.contino.GithubAPI
 def type = "nodejs"
 def product = "civil"
 def component = "citizen-ui"
-def camundaBranch = "master"
+def camundaBranch = "feat/CIV-10391"
```
TatsyMaslova, line 10: these are for testing so even when we approve the pr, you would have to change the branches to master. perhaps its better to create a separate branch from this one just for testing

TatsyMaslova (approved):

hmcts-version1-domini (approved):

PR #2317, opened 2023-10-11
CIV-10998 auth bypass

GorilaNaranja:
### JIRA link ###
https://tools.hmcts.net/jira/browse/CIV-10998

### Change description ###
ITHC BUG - Authentication Bypass

```
[ ] Yes
[x] No
```


src/main/routes/features/public/firstContact/claimSummaryController.ts
```diff
@@ -19,7 +21,11 @@ firstContactClaimSummaryController.get(FIRST_CONTACT_CLAIM_SUMMARY_URL,
   async (req: Request, res: Response, next: NextFunction) => {
     try {
       const cookie = req.cookies['firstContact'];
-      if (cookie?.claimId && cookie?.pinVerified === YesNo.YES) {
+
+      var bytes  = CryptoJS.AES.decrypt(cookie?.hash, cookie.claimReference);
```
sampankumar, line 25: looks like you are using claimReference as the key.
I would suggest use the pin. 
Claim reference would be a wider knowledge

GorilaNaranja: We are not saving the pin anywhere so it won't be able to decrypt in the next page.

GorilaNaranja: I have to do some changes because we didn't have that model so:
- I added pin field to the claim model (claim.respondent1PinToPostLRspec.accessCode).
- I modified the claimSummaryController adding some conditions to control that the cookie and the claim has needed fields.
- Fixed tests

jarekPierchala (approved):

sampankumar (approved):

PR #2381, opened 2023-10-23
CIV-10849 dashboard-design-structure

LeonardoPalmeiro:
### JIRA link ###

https://tools.hmcts.net/jira/browse/CIV-10849

### Change description ###

```
[ ] Yes
[x] No
```


jarekPierchala (changes requested):
impacts R1 - should we expect that?

LeonardoPalmeiro:
> impacts R1 - should we expect that?

It won't mess with R1, but it'll mess with the release 2 tickets. Keep in mind that the first version is just placeholders, and there's no actual functionality in there, so responding to claims, for example, won't work.

For your peace of mind, we've created a branch called "dashboard_dev," and you can rest assured that it won't disrupt any aspect of the CUI.

src/main/common/models/taskList/TaskStatus.ts
```diff
@@ -1,4 +1,17 @@
...
   COMPLETE = 'COMPLETE',
   INCOMPLETE = 'INCOMPLETE',
+  DONE = 'DONE',
+  READY_TO_VIEW = 'READY_TO_VIEW',
+  NOT_AVAILABLE_YET = 'NOT_AVAILABLE_YET',
+  IN_PROGRESS = 'IN_PROGRESS',
 }
+
+export const TaskStatusColor: Record<TaskStatus, string> = {
+  [TaskStatus.COMPLETE]: '',
```
maciejmodzelewski, line 11: What is the colour for this task status?

src/main/modules/i18n/locales/cy.json
```diff
@@ -2437,7 +2439,11 @@
       "HEADER": "Ymateb i hawliad ariannol",
       "ALL_ACTIONS_COMPLETED_CHECK_ANSWERS": "Ar ôl i chi gwblhau'r holl gamau byddwch yn mynd ymlaen i dudalen lle gallwch wirio'ch atebion cyn eu cyflwyno.",
       "COMPLETE": "Cyflawn",
-      "INCOMPLETE": "Anghyflawn"
+      "DONE": "DONE",
```
maciejmodzelewski, line 2442: Why this particular entry is capitalised?

src/main/modules/i18n/locales/en.json
```diff
@@ -2437,7 +2439,11 @@
       "HEADER": "Respond to a money claim",
       "ALL_ACTIONS_COMPLETED_CHECK_ANSWERS": "After you have completed all the actions you will be taken to a page where you can check your answers before submitting.",
       "COMPLETE": "Complete",
-      "INCOMPLETE": "Incomplete"
+      "DONE": "DONE",
```
maciejmodzelewski, line 2442: Why this particular entry is capitalised?

src/main/common/models/taskList/TaskStatus.ts
LeonardoPalmeiro: without specific colour as I told this is just a place holder and this code had been merged to master then revert because broke all CUI Release 2 tests.

maciejmodzelewski: But if it doesn't have any specific colour assigned then is there some default colour that will be applied or the code is going to break when this status is used?

GorilaNaranja: If we don't specify a color it will use the default color of the tag component, and the default color is the correct one for this status.

src/main/modules/i18n/locales/cy.json
GorilaNaranja: Good point, my bad. It should be "Done"

src/main/modules/i18n/locales/en.json
GorilaNaranja: Good point, my bad. It should be "Done"

maciejmodzelewski (approved):

PR #2396, opened 2023-10-25
CIV-11152 Fix Sign settlement agreement content

GorilaNaranja:
### JIRA link ###
https://tools.hmcts.net/jira/browse/CIV-11152

### Change description ###
Admit all & set date content fix

**Does this PR introduce a breaking change?** (check one with "x")

```
[ ] Yes
[x] No
```

PR #2470, opened 2023-11-02
CIV-8027 respond settlement agreement confirmation

paul-pearson:
### JIRA link ###

https://tools.hmcts.net/jira/browse/CIV-8027

### Change description ###

Implement respond settlement agreement confirmation

```
[ ] Yes
[x] No
```


src/main/common/utils/repaymentUtils.ts
```diff
@@ -17,13 +18,13 @@ export const getFinalPaymentDate = (claim: Claim): Date => {
 
   switch (repaymentFrequency) {
     case TransactionSchedule.WEEK:
-      finalRepaymentDate = addDaysToDate(firstRepaymentDate, (numberOfInstalments * WEEKDAYS));
+      finalRepaymentDate = addDaysToDate(firstRepaymentDate, (numberOfInstalments - 1) * WEEKDAYS);
```
portizdev, line 21: We use this function in 5 different controllers/services. Have you checked that this does not impact other tasks that are already completed? Even some of the R1?

src/main/common/utils/repaymentUtils.ts
```diff
@@ -59,6 +60,12 @@ export const getFirstRepaymentDate = (claim: Claim): Date => {
   return new Date(claim.partialAdmission?.paymentIntention?.repaymentPlan?.firstRepaymentDate);
 };
 
+export const getPaymentOptionType = (claim: Claim): PaymentOptionType => {
```
portizdev, line 63: You can use getPaymentIntention() function from claim.ts

src/main/routes/features/settlementAgreement/respondSettlementAgreementConfirmationController.ts
```diff
@@ -0,0 +1,27 @@
...
+const respondSettlementAgreementConfirmationController = Router();
+
+respondSettlementAgreementConfirmationController.get(DEFENDANT_SIGN_SETTLEMENT_AGREEMENT_CONFIRMATION, respondSettlementAgreementConfirmationGuard, (async (req, res, next) => {
+  try {
+    const lang = req.query.lang ? req.query.lang : req.cookies.lang;
+    const claimId = req.params.id;
+    const claim = await getClaimById(claimId, req, true);
+    claim.id = claimId;
+    const claimantResponseConfirmationContent = getRespondSettlementAgreementConfirmationContent(claim, getLng(lang));
+    res.render('features/settlementAgreement/respond-settlement-agreement-confirmation', {
```
portizdev, line 19: Use const for the view path

src/main/routes/features/settlementAgreement/respondSettlementAgreementController.ts
```diff
@@ -0,0 +1,83 @@
...
+import {AppRequest} from 'models/AppRequest';
+
+const respondSettlementAgreementViewPath = 'features/settlementAgreement/respond-settlement-agreement';
+const respondSettlementAgreementController = Router();
+
+function renderView(form: GenericForm<GenericYesNo>, res: Response, data?: object): void {
+  res.render(respondSettlementAgreementViewPath, {form, data});
+}
+
+const getSettlementAgreementData = (claim: Claim, req: Request) => {
```
portizdev, line 29: Create a separate service for this functions. We need to avoid logic in the controllers

src/main/routes/features/settlementAgreement/respondSettlementAgreementController.ts
```diff
@@ -0,0 +1,83 @@
...
+  return data;
+};
+
+respondSettlementAgreementController.get(DEFENDANT_SIGN_SETTLEMENT_AGREEMENT, (async (req: Request, res: Response, next: NextFunction) => {
+  try {
+    const claimId = req.params.id;
+    const claim = await getClaimById(claimId, req, true);
+    let selectedOption: string;
+    if (claim.defendantSignedSettlementAgreement) {
+      selectedOption = 'yes';
```
portizdev, line 52: We have constants for this strings. And pls use a separate service for this logic.

src/main/routes/features/settlementAgreement/respondSettlementAgreementController.ts
```diff
@@ -0,0 +1,83 @@
...
+  try {
+    const claimId = req.params.id;
+    const respondSettlementAgreement = new GenericForm(new GenericYesNo(req.body.option, 'PAGES.DEFENDANT_RESPOND_TO_SETTLEMENT_AGREEMENT.DETAILS.VALID_YES_NO_OPTION'));
+    respondSettlementAgreement.validateSync();
+
+    const claim = await getClaimById(claimId, req);
+    if (respondSettlementAgreement.hasErrors()) {
+      renderView(respondSettlementAgreement, res, getSettlementAgreementData(claim, req));
+    } else {
+      claim.defendantSignedSettlementAgreement = respondSettlementAgreement.model.option === 'yes';
```
portizdev, line 72: Why do you use a different model for the claim and for the view? We should only send an object that is encapsulated in a GenericForm and use that model and retrieve it directly without the need to do conversions between models.

src/main/views/features/settlementAgreement/respond-settlement-agreement-confirmation.njk
```diff
@@ -0,0 +1,33 @@
+{% extends "claim-details-tpl.njk" %}
+{% from "govuk/components/panel/macro.njk" import govukPanel %}
```
portizdev, line 2: You don't need to import it if you don't use it

src/main/views/features/settlementAgreement/respond-settlement-agreement.njk
```diff
@@ -0,0 +1,96 @@
...
+{% block pageTitle %}
+  {{ setPageTitle('PAGES.DEFENDANT_RESPOND_TO_SETTLEMENT_AGREEMENT.PAGE_TITLE') }}
+{% endblock %}
+{% set yesHtml %}
+  <p class="govuk-body">{{ t('PAGES.DEFENDANT_RESPOND_TO_SETTLEMENT_AGREEMENT.DETAILS.ON_HOLD') }}</p>
+{% endset %}
+{% set noHtml %}
+  <p class="govuk-body">{{ t('PAGES.DEFENDANT_RESPOND_TO_SETTLEMENT_AGREEMENT.DETAILS.REQUEST_CCJ') }}</p>
+{% endset %}
+{% if data.repaymentFrequency === TransactionSchedule.WEEK %}
```
portizdev, line 19: We can't use so much logic in the view, we must send the texts correctly from the backend. You can use this PR ( #2396  ) as an example where we have modified this file by doing all this logic in the backend.

src/test/unit/routes/features/settlementAgreement/respondSettlementAgreementConfirmationController.test.ts
```diff
@@ -0,0 +1,57 @@
...
+  function getMockClaim() {
+    const mockClaim = new Claim();
+    mockClaim.fullAdmission = new FullAdmission();
+    mockClaim.fullAdmission.paymentIntention = new PaymentIntention();
+    mockClaim.fullAdmission.paymentIntention.paymentOption = PaymentOptionType.BY_SET_DATE;
+    return mockClaim;
+  }
+
+  describe('on GET', () => {
+    it('should return accept settlement agreement confirmation', async () => {
```
portizdev, line 32: Use Given - When - Then pattern pls

portizdev (changes requested):

src/main/common/utils/repaymentUtils.ts
paul-pearson: Are you suggesting that this isn't a bug after all?  Even if this is used in R1, I believe it is still a bug that is now fixed.  Also this change is part of CIV-8026 which is already in PO review.

src/main/routes/features/settlementAgreement/respondSettlementAgreementConfirmationController.ts
paul-pearson: Done

src/main/views/features/settlementAgreement/respond-settlement-agreement-confirmation.njk
paul-pearson: Removed.

src/test/unit/routes/features/settlementAgreement/respondSettlementAgreementConfirmationController.test.ts
paul-pearson: Done

src/main/views/features/settlementAgreement/respond-settlement-agreement.njk
paul-pearson: Discussed in the PR for 8026 - now in PO review.  Agreed this logic would have been better not in the view, but was done for consistency with the existing claimant settlement agreement screen.  If refactored, that existing screen should be refactored similarly.

src/main/routes/features/settlementAgreement/respondSettlementAgreementController.ts
paul-pearson: Replacing the claim model to match the form has resolved this.

src/main/routes/features/settlementAgreement/respondSettlementAgreementController.ts
paul-pearson: Replaced the claim model to match the form model

src/main/routes/features/settlementAgreement/respondSettlementAgreementController.ts
paul-pearson: Agreed this could be moved, however was done this way for consistency with existing sign settlement agreement controller for claimant.  If we refactor here, the existing controller should also be refactored. This is another change from 8026 which is already in PO approval.

src/main/common/utils/repaymentUtils.ts
portizdev: It may solve the bug on the page you are working on but when looking at 4 other sites where it is used you should check that it has not affected those other views/calculations that are used in R1.

paul-pearson: This has been checked.  The function is not used in R1.  In R2 the original bug is on a number of other pages, which are now fixed by this change.

portizdev (dismissed):
Super work Paul! Congrats 🎉

ElifDenizEsea (dismissed):

neeta-hmcts (dismissed):

PR #2605, opened 2023-11-20
CIV-8454 Generate service request

paul-pearson:
### JIRA link ###

https://tools.hmcts.net/jira/browse/CIV-8454

### Change description ###

Fire event to trigger service request creation

```
[ ] Yes
[x] No
```


sonar-project.properties
```diff
@@ -49,7 +49,8 @@ src/main/routes/features/claimantResponse/partPaymentReceivedController.ts, \
 src/main/routes/features/claimantResponse/ccj/ccjConfirmationController.ts, \
 src/main/routes/features/directionsQuestionnaire/supportRequiredController.ts, \
 src/main/routes/features/directionsQuestionnaire/witnesses/otherWitnessesController.ts, \
-src/main/assets/js/fileValidator.js
+src/main/assets/js/fileValidator.js, \
+src/main/routes/features/claim/totalAmountController.ts
```
laarmada, line 53: which is the issue to skip sonnar? totalAmountController

src/main/services/features/claim/amount/claimFeesService.ts
```diff
@@ -1,13 +1,17 @@
...
 
 const {Logger} = require('@hmcts/nodejs-logging');
 const logger = Logger.getLogger('claimantPhoneAsService');
-export const saveClaimFee = async (claimantId: string, claimFee: number) => {
+export const saveClaimFee = async (claimantId: string, claimFeeData: ClaimFeeData) => {
   try{
     const claim = await getCaseDataFromStore(claimantId);
     console.log(claimantId);
-    claim.claimFee = {calculatedAmountInPence: convertToPence(claimFee)};
+    claim.claimFee = {
```
laarmada, line 10: only a question which journey is this claimFee?

dharmendrak (approved):

src/main/app/client/civilServiceClient.ts
```diff
@@ -155,11 +155,16 @@ export class CivilServiceClient {
...
+    const claimFeeData = await this.getClaimFeeData(amount, req);
+    return convertToPoundsFilter(claimFeeData?.calculatedAmountInPence.toString());
+  }
+
+  async getClaimFeeData(amount: number, req: AppRequest): Promise<ClaimFeeData> {
     const config = this.getConfig(req);
     try {
       const response: AxiosResponse<object> = await this.client.get(`${CIVIL_SERVICE_CLAIM_AMOUNT_URL}/${amount}`, config);
       const claimFeeResponse: ClaimFeeData = response.data;
-      return convertToPoundsFilter(claimFeeResponse?.calculatedAmountInPence.toString());
```
laarmada, line 162: really good to see that removed once is adding in civil-service 
image

laarmada (commented):
ok

jeswanth-hmcts (approved):

sonar-project.properties
paul-pearson: All the claim tests are disabled in jest here: https://github.com/hmcts/civil-citizen-ui/blob/2275b1f95c92c268ad714e675f53d018015d5d1b/jest.config.js#L5
So this means I can't get test coverage for my changes in totalAmountController.ts.
I think @portizdev has been working on fixing the broken tests which will allow us to re-enable all these tests and remove sonar exclusions.

src/main/services/features/claim/amount/claimFeesService.ts
paul-pearson: It is the same journey - claim creation.  On screen /claim/total we calculate the fee from civil service request.  Previously we were only saving the fee calculatedAmountInPence.  After this change we will be saving the additional fields so we can use them in service request creation.

sonar-project.properties
laarmada: yes but please don't add more files in the exclusions... we need to figure out the reason and try to fix it. That works is included in the story points estimations.

src/main/services/features/claim/amount/claimFeesService.ts
laarmada: we should stop to add isolated vbles in claim directly we need to create hierarchy in order to know in which journey is used that data.... else we will have the same problem as CCD

laarmada (approved):

PR #3168, opened 2024-02-19
CIV-12060 Prevent creation of DJ Spec form for LR V LiP claims

JackDevN:
### JIRA link ###

https://tools.hmcts.net/jira/browse/CIV-12060

### Change description ###

The scope of this ticket is to prevent the generation of the order form (see attached document for example) for default judgments in LR V LiP claims, ensure that this form is not being sent to any actors in the journey (Claimant, defendant or case worker), and to ensure that this change bears no impact on the LR V LR journey, where the form is generated.

This change in civil-citizen-ui should fix any broken logic made from the changes from the CIV-12060 to the civil-service side.

```
[ ] Yes
[ ] No
```


src/main/common/models/claim.ts
```diff
@@ -800,7 +802,7 @@ export class Claim {
   }
 
   hasClaimTakenOffline() {
-    return this.ccdState === CaseState.PROCEEDS_IN_HERITAGE_SYSTEM && !this.defaultJudgmentDocuments && !this.ccjJudgmentStatement && !this.isClaimantRejectedPaymentPlan();
+    return this.ccdState === CaseState.PROCEEDS_IN_HERITAGE_SYSTEM && !this.ccjJudgmentStatement && !this.isClaimantRejectedPaymentPlan();
```
asthamalviya, line 805: Can you please confirm with Domini but as far as I remember Neeta said we will need to add a check here too.

hmcts-version1-domini (approved):

asthamalviya (approved):

PR #3273, opened 2024-03-05
CIV-1511 Add different party types scenarios for claim creation

mounikahmcts:


src/test/functionalTests/codecept.conf.js
```diff
@@ -13,7 +13,7 @@ exports.config = {
     await deleteAllIdamTestUsers();
   },
 
-  tests: '../functionalTests/tests/**/*_tests.js',
+  tests: '../functionalTests/tests/**/*.js',
```
rajakm, line 16: need to revert this

PR #3517, opened 2024-04-09
 CIV-0000 Add common tpl file to cover bread crum and spacing issue

LeonardoPalmeiro:

### Change description ###

```
[ ] Yes
[ ] No
```


src/main/views/claim-details-tpl-dashboard.njk
```diff
@@ -0,0 +1,153 @@
...
+        tag: {
+          text: "beta"
+        },
+        html: bannerText + languageToggle
+      }) }}
+
+{% if backLinkUrl %}
+  {{ back(backLinkUrl, t, false, false) }}
+{% else %}
+{{ govukBreadcrumbs({
```
m-meulendijks-v1, line 52: Are you sure the breadcrumbs will always point to /Dashboard?

I ask, because several pages actually seem to point to the claim summary pages, and the language on the button of most pages 'return to case overview' suggests the claim summary of 1 case, rather than the Dashboard for all cases.

Consequently, the page will be different for claimant/defendant.

LeonardoPalmeiro: I'm not sure honesty .
Might be an good option to ask to BI's to make sure where should we point it.
I just add the component.

m-meulendijks-v1: Sure, I've changed it to accept different URLs for my PR.
But you don't necessarily need to change it for yours.

PR #3757, opened 2024-05-09
CIV-13650 - reformatting Dashboard page

m-meulendijks-v1:
### JIRA link ###
https://tools.hmcts.net/jira/browse/CIV-13650

### Change description ###
image
image

```
[ ] Yes
[x] No
```


miguelMolina3691 (approved):

src/main/views/features/dashboard/claim-summary-redesign.njk
```diff
@@ -1,5 +1,6 @@
-{% extends "claim-details-tpl.njk" %}
+{% extends "claim-details-tpl-dashboard.njk" %}
 {% from "govuk/components/tabs/macro.njk" import govukTabs %}
+{% from "govuk/components/breadcrumbs/macro.njk" import govukBreadcrumbs %}
```
laarmada, line 3: Please remove the unnecessary imports

src/main/views/features/dashboard/claim-summary-redesign.njk
```diff
@@ -12,25 +13,29 @@
 
   <div class="govuk-width-container govuk-!-padding-top-0">
```
laarmada, line 14: if you use the new claim-details-tpl-dashboard.njk template we can remove that div lines

src/main/views/features/dashboard/claim-summary-redesign.njk
```diff
@@ -12,25 +13,29 @@
...
       <div class="govuk-grid-row">
         {# Left side #}
         <div class="govuk-grid-column-two-thirds">
           <div class="border-small-top"></div>
             {% for notification in dashboardNotifications.items %}
                 {{ govukNotificationBanner({
                    titleText: notification.titleCy if lang=="cy" else notification.titleEn,
-                   html: notification.descriptionCy if lang=="cy" else notification.descriptionEn
+                   html: notification.descriptionCy if lang=="cy" else notification.descriptionEn,
+                   classes: "govuk-!-margin-bottom-4"
```
laarmada, line 38: we can remove the contact us for help macro, exist in the new tamplate.

src/main/views/features/dashboard/claim-summary-redesign.njk
```diff
@@ -12,25 +13,29 @@
...
-        <div class="govuk-grid-column-two-thirds govuk-!-margin-top-0">
-          <h1 class="govuk-heading-m govuk-!-margin-0">{{ claimantName }} v {{ defendantName }}</h1>
-          <p class="govuk-body-s">
-              {{ t('PAGES.CLAIM_SUMMARY.CLAIM_NUMBER') }}
-              {{ claim.legacyCaseReference }}
-          </p>
+      {% if claimId !== 'draft' %}
+        {% set claimantName = claim.getClaimantFullName() %}
+        {% set defendantName = claim.getDefendantFullName() %}
+        <div class="govuk-grid-row govuk-!-margin-top-0">
```
laarmada, line 19: we can remove every div with style classes.

src/main/views/features/dashboard/claim-summary-redesign.njk
laarmada: review also if the right "support" section exist in the current template or create a reusable macro.

laarmada (changes requested):

src/main/views/features/dashboard/claim-summary-redesign.njk
m-meulendijks-v1: I am the 'content' block rather than the 'nestedContent', because otherwise the 1-3rd column won't contain the information it needs. I will detail that more in your other comment. This also means I need to keep these divs, because I overwrite the ones from the template.

src/main/views/features/dashboard/claim-summary-redesign.njk
m-meulendijks-v1: It is different in the new template. The new template looks like this:
```
{% block content %}
<div class="govuk-grid-row">
  <div class="govuk-grid-column-two-thirds">
    {% block nestedContent %}{% endblock %}
  </div>
  <div class="govuk-grid-column-one-third">
    {{ contactUsForHelp(t) }}
  </div>
</div>
{% endblock %}
```
But as you can see, that means the second column only contains the contactUsForHelp, while there needs to be more information in the column like so:
```
{# TODO: Right side #}
        <div class="govuk-grid-column-one-third">
          <div class="border-small-top"></div>
          {% if iWantToLinks.length %}
            <div class="govuk-!-margin-bottom-8">
              {{ supportLinks(t, iWantToTitle, iWantToLinks) }}
            </div>
          {% endif %}
          {% if helpSupportLinks.length %}
            {{ supportLinks(t, helpSupportTitle, helpSupportLinks) }}
          {% endif %}
          <div class="govuk-!-padding-top-8">
            {{ contactUsForHelp(t) }}
          </div>
        </div>
```
So I need to overwrite the the content block from the template. I am still using the new template in case other elements on the template get changed, it is better that all pages use a similar template rather than some using one, and others using another.

src/main/views/features/dashboard/claim-summary-redesign.njk
m-meulendijks-v1: Removed several unnecessary imports, many thanks.

src/main/views/features/dashboard/claim-summary-redesign.njk
m-meulendijks-v1: I've removed two div lines that could indeed be removed.

src/main/views/features/dashboard/claim-summary-redesign.njk
laarmada: Could we move that block of code to a Macro and reuse in other pages? or not make sense?

laarmada (approved):

miguelMolina3691 (approved):

src/main/views/features/dashboard/claim-summary-redesign.njk
m-meulendijks-v1: We could do it, but with the rush behind this ticket I'm not keen to do it, right now.
If another ticket for dashboard pops up, I will try and put this in a macro.

dharmendrak (approved):

PR #3777, opened 2024-05-13
CIV-9442 Implement application costs page

paul-pearson:
### JIRA link ###

https://tools.hmcts.net/jira/browse/CIV-9442

### Change description ###

Implement application costs page

```
[ ] Yes
[x] No
```


src/main/app/client/civilServiceClient.ts
```diff
@@ -191,6 +193,31 @@ export class CivilServiceClient {
     }
   }
 
+  async getGeneralApplicationFee(applicationTypeOption: string, withConsent: YesNo, withNotice: YesNo, req: AppRequest): Promise<number> {
+    const gaFeeData = await this.getGeneralApplicationFeeData(applicationTypeOption, withConsent, withNotice, req);
```
neeta-hmcts, line 197: we do not need to save this gaFeeData somewhere in the claim object.

neeta-hmcts (approved):

src/main/app/client/civilServiceClient.ts
paul-pearson: As discussed, I don't think we need to save this because it is not the final fee which the citizen will pay.  They may choose to add another application, or the fee may be free if the hearing date is more than 14 days away, for example.

manish14836 (approved):

PR #4719, opened 2024-10-08
DTSCCI-843 Remove banner from CUI following CUI R2

tiagreaney:
### JIRA link ###
https://tools.hmcts.net/jira/browse/DTSCCI-834


src/main/assets/scss/main.scss
```diff
@@ -303,60 +303,3 @@
   margin-top: 10px;
   padding: 10px;
 }
-
```
gmikerlemos, line 306: Shouldn't you just remove the entry in the template as these banners are used frequently?

src/main/views/macro/banner.njk
```diff
@@ -1,12 +0,0 @@
-<div class="hmcts-banner">
```
gmikerlemos, line 1: Thinking whether you should keep it to be reused later when we need other banners

gmikerlemos (approved):

velimd (approved):

nigeldunne (approved):

PR #4856, opened 2024-10-22
CIV-15627 Save user id and original url to redis

paul-pearson:
### JIRA link ###

https://tools.hmcts.net/jira/browse/CIV-15627

### Change description ###

Fix for duplicate payment issue, using redis

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/modules/oidc/index.ts
+++ b/src/main/modules/oidc/index.ts
@@ -109,9 +109,12 @@ export class OidcMiddleware {
           app.locals.claimIssueTasklist = undefined;
           return res.redirect(CLAIMANT_TASK_LIST_URL);
         }
-        if (app.locals.paymentConfirmationUrl) {
-          const paymentConfirmationUrl = app.locals.paymentConfirmationUrl;
-          app.locals.paymentConfirmationUrl = '';
+        const draftStoreClient = app.locals.draftStoreClient;
+        logger.info('login user id ', req.session.user.id);
+        const paymentConfirmationUrl = await draftStoreClient.get(req.session.user.id + 'userIdForPayment');
+        logger.info('Payment conf url ', paymentConfirmationUrl);
+        if (paymentConfirmationUrl) {
+          await draftStoreClient.del(req.session.user.id + 'userIdForPayment');
           return res.redirect(paymentConfirmationUrl);
         }
         if (req.session.user?.roles?.includes(citizenRole)) {
@@ -139,7 +142,7 @@ export class OidcMiddleware {
       res.redirect(DASHBOARD_URL);
     });
 
-    app.use((req: Request, res: Response, next: NextFunction) => {
+    app.use(async (req: Request, res: Response, next: NextFunction) => {
       const appReq: AppRequest = <AppRequest>req;
       if (appReq.session?.user) {
         if (appReq.session.user.roles?.includes(citizenRole)) {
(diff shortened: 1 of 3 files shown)


src/main/modules/oidc/index.ts
```diff
@@ -139,38 +147,63 @@ export class OidcMiddleware {
...
         }
+        if (requestIsForAssigningClaimForDefendant(req)) {
+          app.locals.assignClaimURL = appReq.session.assignClaimURL = ASSIGN_CLAIM_URL;
+        }
+        if (requestIsForClaimIssueTaskList(req)) {
+          app.locals.claimIssueTasklist = appReq.session.claimIssueTasklist = true;
+        }
+        logger.info('redirecting url ', req.originalUrl);
+        if (isPaymentConfirmationUrl(req)) {
+          const claimId = getClaimId(req.originalUrl);
```
paul-pearson, line 179: I think we should add some checking here to make sure claimId is not undefined

src/main/modules/oidc/index.ts
```diff
@@ -139,38 +147,63 @@ export class OidcMiddleware {
...
+          logger.info('Condition satisfied for payment confirmation ', req.originalUrl);
+          logger.info('Claim id ', claimId);
+          const userId = await getUserId(claimId, app);
```
paul-pearson, line 182: As above, check for undefined value so that we don't end up saving with a key something like "undefineduserIdForPayment".

src/main/modules/oidc/index.ts
```diff
@@ -139,38 +147,63 @@ export class OidcMiddleware {
...
+export const getUserId = async (claimId: string, app: Application): Promise<string> => {
+  return await app.locals.draftStoreClient.get(claimId + 'userIdForPayment');
+};
+
+export const getPaymentConfirmationUrl = async (userId: string, app: Application): Promise<string> => {
+  return await app.locals.draftStoreClient.get(userId + 'userIdForPayment');
+};
+
+const getClaimId = (originalUrl: string) => {
+  return originalUrl.split('/')[2];
```
paul-pearson, line 208: Add some checking here to make sure split array has sufficient length and maybe return null/undefined if not (we can check for this after return).  Could also check claimId is as expected (16 digit).

src/main/modules/oidc/index.ts
jeswanth-hmcts: so in case of  claim ID is undefined , log the error bypassing the save

src/main/modules/oidc/index.ts
jeswanth-hmcts: added extra check

neeta-hmcts (approved):

src/main/modules/oidc/index.ts
```diff
@@ -139,38 +147,70 @@ export class OidcMiddleware {
...
+export const getUserId = async (claimId: string, app: Application): Promise<string> => {
+  return await app.locals.draftStoreClient.get(claimId + 'userIdForPayment');
+};
+
+export const getPaymentConfirmationUrl = async (userId: string, app: Application): Promise<string> => {
+  return await app.locals.draftStoreClient.get(userId + 'userIdForPayment');
+};
+
+const getClaimId = (originalUrl: string) => {
+  const extractClaimId = originalUrl?.split('/')[2];
```
paul-pearson, line 212: Here to be safe I would have: 
```
const originalUrlArray =originalUrl?.split('/');
if (originalUrlArray?.length >=3) {
  const extractClaimId = originalUrlArray[2];
  if (extractClaimId && extractClaimId.length === 16) {
    return extractClaimId;
  }
}
```
or similar.

paul-pearson: Discussed with Jeswanth - this check is not required.

paul-pearson (commented):
PR approved.  I am only able to leave comments as I created the PR.

dharmendrak (approved):

LeonardoPalmeiro (approved):

PR #5035, opened 2024-11-13
DTSCCI-1149 Round value to avoid floating point error

velimd:
### JIRA link ###
https://tools.hmcts.net/jira/browse/DTSCCI-1149

### Change description ###
Fix to solve the issue where some values when multipled by 100 causes a floating point error.

```
[ ] Yes
[X] No
```


rishikrsharma (dismissed):

AbuSalamHMCTS (approved):

src/main/services/translation/response/ccdTranslation.ts
```diff
@@ -67,7 +67,7 @@ export const translateDraftResponseToCCD = (claim: Claim, addressHasChange: bool
     respondToAdmittedClaim: toCCDRespondToClaim(claim.partialAdmission?.howMuchHaveYouPaid),
     specDefenceAdmittedRequired: toCCDYesNoFromGenericYesNo(claim.partialAdmission?.alreadyPaid),
     respondToAdmittedClaimOwingAmountPounds: claim.partialAdmission?.howMuchDoYouOwe?.amount?.toString(),
-    respondToAdmittedClaimOwingAmount: claim.partialAdmission?.howMuchDoYouOwe?.amount ? (claim.partialAdmission?.howMuchDoYouOwe?.amount * 100).toString(): undefined,
+    respondToAdmittedClaimOwingAmount: claim.partialAdmission?.howMuchDoYouOwe?.amount ? Math.round((claim.partialAdmission?.howMuchDoYouOwe?.amount * 100)).toString(): undefined,
```
rishikrsharma, line 70: seems like it will round - we need to scale to 2 point.
let number = 5.7;
let rounded = Math.round(number);

console.log(rounded); // Outputs: 6

velimd: Just to mention again here incase anyone else reviews it. The value is already to 2 decimal points as its for the pennies, e.g. howMuchDoYouOwe?.amount = 123.45. We have another variables which multplies it by 100 which i assume to get it in pennies or as a whole number so 123.45 * 100 would be 12345 however, some values would cause a floating point error e.g. 123.45 * 100 gives 12344.9999999 hence I use Math.round.

nigeldunne (approved):

rishikrsharma (approved):

tiagreaney (approved):

PR #5076, opened 2024-11-19
CIV-5659 - Confirm your payment screen not translated to Welsh

neeta-hmcts:
### JIRA link ###
https://tools.hmcts.net/jira/browse/CIV-15659

### Change description ###
Added code changes to toggle a payment successful page based on the language selected.

```
[ ] Yes
[x] No
```


src/main/services/features/generalApplication/payment/applicationFeePaymentConfirmationService.ts
```diff
@@ -1,41 +1,48 @@
...
 
 const {Logger} = require('@hmcts/nodejs-logging');
 const logger = Logger.getLogger('applicationFeePaymentConfirmationService');
 
 const success = 'Success';
 const paymentCancelledByUser = 'Payment was cancelled by the user';
 
+function getApplicantApplicationIssueLang(claim: Claim) {
+  if(claim.isClaimant()) {
+    return claim.claimantBilingualLanguagePreference === ClaimBilingualLanguagePreference.WELSH_AND_ENGLISH ? 'cy' : 'en';
```
jeswanth-hmcts, line 23: applicantBilingualLanguagePreference,respondentBilingualLanguagePreference, don't  we need to consider this ccd property instead of the main claim,  assuming that  these values will be set during application submission

neeta-hmcts: Yes... Those fields can also be used instead as both will give you the same result.  
I find this way more readable hence used this logic. In the end, we use claim values to set the applicantBilingualLanguagePreference ,respondentBilingualLanguagePreference.

Do you find any issues with this logic ?

jeswanth-hmcts: suppose there is a scenario where the defendant created an application before responding at this time the flag  is set to no for  respondentBilingualLanguagePreference and he paid the fee after the  submission of the defendant's response in Welsh and this page will show in Welsh instead of English, it should consist with other pages since  respondentBilingualLanguagePreference is set to be no

neeta-hmcts: Yes.. I agree with you. I will make the changes. Thanks for pointing out my mistake.

jeswanth-hmcts (dismissed):

jeswanth-hmcts (approved):

paul-pearson (approved):

PR #5864, opened 2025-03-27
Civ 16704 send query form

kurt-uk:
### JIRA link ###
https://tools.hmcts.net/jira/browse/CIV-16704

### Change description ###
Added send query form, Added start point from what you want to do questions "send a message" link text.
Created controllers to handle page and form management with file uploader to civil service and then a new redis key store for the saved case documents to be retrieved when needed.
Added validation
renamed directories from `qm` -> `queryManagement` for clarity
## Screenshots
image
image
image
image

```
[ ] Yes
[ ] No
```


src/main/routes/features/queryManagement/createQueryController.ts
```diff
@@ -0,0 +1,110 @@
...
+  const formattedSummary = summarySection(
+    {
+      title: '',
+      summaryRows: [],
+    });
+  if (req.query?.id) {
+    const index = req.query.id;
+    await removeSelectedDocument(redisKey, Number(index)-1);
+  }
+  if (req?.session?.fileUpload) {
```
jeswanth-hmcts, line 67: The data related to fields such as subject, title, and hearing is currently stored in session Redis under fileupload, while uploaded document files are stored in local Redis. This approach may lead to inconsistencies for the user. For instance, when the session expires or the user logs out, the data for fields subject, title, and hearing will be lost. However, the user can still see the uploaded files on the same page upon returning.
To ensure consistency, the entire page should maintain a unified approach: either all data should be visible, or none of it should appear upon returning. Additionally, this solution should align with other query management-related pages where data is stored in app.local Redis. It is advisable that all data for a page be stored under a single related object within the model. Otherwise, referencing data from multiple places across different pages becomes cumbersome, as is the current situation where fields such as subject, title, and hearing are stored under property, while uploaded files are stored in app.local Redis with different Redis keys.

drummondjm (dismissed):

jeswanth-hmcts (approved):

LeonardoPalmeiro (approved):

drummondjm (approved):

jeswanth-hmcts (approved):

LeonardoPalmeiro (approved):

PR #7806, opened 2026-06-17
DTSCCI-5282: surface civil-service 422 callback errors 

ruban72:
...instead of generic 500

When civil-service returns 422 with a callbackErrors body for a rejected citizen-event (e.g. the businessProcess guard), CUI previously treated it as a generic failure and rendered the "Something went wrong" page, discarding the user-actionable message.

submitEvent now throws a typed CallbackError carrying the callbackErrors, and the global error middleware renders those messages at HTTP 422 (no App Insights exception, as it is a 4xx). Centralised at the single submitEvent chokepoint so it covers all affected CYA routes (qm, claimant/defendant response, paid-in-full, GA, request-for-reconsideration, etc.). Happy path is unchanged.

Add unit tests for submitEvent (422 -> CallbackError; re-throw otherwise) and the error middleware rendering path.

### JIRA link ###
[DTSCCI-5282](https://tools.hmcts.net/jira/browse/DTSCCI-5282)

### Change description ###
surface civil-service 422 callback errors


src/main/modules/error/index.ts
```diff
@@ -18,6 +19,19 @@ export class ErrorHandler {
       const errorMessage = err.message || 'Internal Server Error';
       const status = (err as HTTPError)?.status || 500;
       logger.error(`${err.stack || errorMessage}`);
+
+      // DTSCCI-5282: a rejected citizen-event callback (422) carries user-actionable
+      // messages from CCD. Render those instead of the generic "Something went wrong" page.
+      const callbackError = err as CallbackError;
+      if (callbackError?.callbackErrors?.length) {
```
gmikerlemos: Could we check err instanceof CallbackError here as well? As written, any error with a callbackErrors property will take this path, which could accidentally hide an unrelated 5xx from App Insights and present it as a 422.

src/main/views/error.njk
```diff
@@ -1,7 +1,14 @@
 {% extends "claim-details-tpl.njk" %}
 
 {% block content %}
-  <h1 class="govuk-heading-xl">{{ t('ERRORS.SOMETHING_WENT_WRONG') }}</h1>
-  <p class="govuk-body">{{ t('COMMON.TECHNICAL_PROBLEMS') }}</p>
-  <p class="govuk-body">{{ error.message }}</p>
+  {% if callbackErrors and callbackErrors.length %}
+    <h1 class="govuk-heading-xl">{{ t('ERRORS.THERE_IS_A_PROBLEM') }}</h1>
+    {% for callbackError in callbackErrors %}
+      <p class="govuk-body">{{ callbackError }}</p>
```
gmikerlemos, line 7: How does the user get back to correct a field validation error from here? For messages from details.field_errors, is it would be better to return them to the originating form with an error summary/inline error? If that is outside this ticket maybe we could at least give the user a clear way to go back and amend their answer?

src/main/modules/error/index.ts
ruban72: Good catch — done in c079cdad83. The 422 branch now guards with `err instanceof CallbackError && err.callbackErrors?.length` (dropped the duck-typed cast), so an unrelated 5xx that happens to carry a `callbackErrors` property is no longer masked as a 422 and still hits the App Insights `trackException` below. `CallbackError` is a real class (`extends HTTPError`, constructed/thrown at `civilServiceClient.ts:485`), so `instanceof` holds on the throw path. Added a regression test asserting a non-`CallbackError` carrying `callbackErrors` renders 500 and is tracked.

src/main/views/error.njk
ruban72: Agreed — returning the user to the originating form with an error summary/inline errors is bigger than this ticket's scope ("surface the 422"), so I've split it into a follow-up: **DTSCCI-5899**. That covers routing `details.field_errors` back to the originating form with a GDS error summary + inline errors, keeping the current page for non-field business-rule `callbackErrors`, with a clear go-back-and-amend fallback where the form can't be resolved. Keeping this PR to just surfacing the messages.

ruban72:
@gmikerlemos thanks for the review — both comments are now addressed:

1. **`error/index.ts` (duck-typed 422 check)** — the branch now guards with `err instanceof CallbackError && err.callbackErrors?.length` (dropped the loose cast), so an unrelated 5xx that happens to carry a `callbackErrors` property is no longer masked as a 422 and still hits the App Insights `trackException`. Added a regression test covering exactly that case (non-`CallbackError` with a `callbackErrors` property → 500 + tracked). Done in c079cdad83.

2. **`error.njk` (recovery from field-validation errors)** — agreed this is bigger than this ticket's "surface the 422" scope, so split into follow-up **DTSCCI-5899** (route `details.field_errors` back to the originating form with a GDS error summary + inline errors; keep the current page for non-field business-rule `callbackErrors`, with a go-back-and-amend fallback). This PR stays scoped to surfacing the messages.

Ready for re-review 🙏

gmikerlemos (approved):
lgtm

PR #7820, opened 2026-06-19
DTSCCI-5684: remove the duplicate calls for calculate interest

elvianihuseyin2:
### JIRA link ###

https://tools.hmcts.net/jira/browse/DTSCCI-5684

### Change description ###
**Summary**
Fixes duplicate POST /fees/claim/calculate-interest calls to civil-service by adding request-scoped memoisation, following the same pattern introduced in DTSCCI-5301 for retrieveClaimDetails and getUserCaseRoles.

**Problem**
Post-DTSCCI-5301 KQL showed ~9,613 duplicate interest-calculation POSTs per week (~9.8% waste). 
Screenshot 2026-06-25 at 16 15 39

On a single page render, multiple code paths call calculateInterestToDate independently with the same claim payload (e.g. claim-details calls it via both getInterestDetails and getTotalAmountWithInterestAndFeesAndFixedCost), firing identical POSTs within the same HTTP request.

**What we changed**
Added calculateInterestRequestCache to AppRequest.locals
Memoised calculateClaimInterest in civilServiceClient using a Promise cache keyed on the interest payload (JSON.stringify(claim))
Threaded req through interestUtils, claimDetailsService, controllers, and CCJ translation so all callers within a request share the cache
Added unit tests verifying duplicate calls within the same request result in a single downstream HTTP call

```
[ ] Yes
[ ] No
```


src/main/app/client/civilServiceClient.ts
```diff
@@ -119,6 +119,18 @@ export class CivilServiceClient {
     return newCache;
   }
 
+  private getCalculateInterestRequestCache(req: AppRequest): Map<string, Promise<number>> {
```
anirudha-deshpande, line 157: why we are caching interest, we suppose to call the metod with every interaction with amount where interest is applicable.

elvianihuseyin2: We're not caching interest across user interactions — only deduplicating identical calls within a single HTTP request.

The cache lives on req.locals, which is scoped to one request and cleared when the response is sent. Every new page load or form submission still calls civil-service to calculate interest.

The issue we're fixing: on one render, several helpers (e.g. getInterestDetails and getTotalAmountWithInterestAndFeesAndFixedCost on claim-details) each call calculateInterestToDate with the same claim payload, so we POST to /fees/claim/calculate-interest twice with identical data. KQL shows ~9,613 of these duplicate POSTs per week.

We still recalculate interest on every interaction where it applies — we just avoid firing the same POST multiple times when nothing has changed within that single request.

anirudha-deshpande (approved):

PR #7878, opened 2026-07-03
Dtscci 5685 guard redis double reads fix 2

elvianihuseyin2:
### JIRA link ###
https://tools.hmcts.net/jira/browse/DTSCCI-5685

### Change description ###
**Problem**
Several routes read the same claim from Redis twice per request — once in a guard (middleware) and again in the downstream controller. This adds unnecessary Redis round-trips on every page load.

The worst case is /response/check-and-send: allResponseTasksCompletedGuard calls getClaimById (which can fall through to civil-service on a cold cache), but the controller then ignored that and read Redis directly via getCaseDataFromStore.

**Solution**
Extended AppRequest.locals with an optional typed claim field
Added claimRequestLocals.ts with:
- stashClaimOnRequest() — stores claim on the request
- getStashedClaimOrFromStore() — returns stashed claim, or falls back to Redis if no guard ran
Guards stash before next(); 
controllers read from locals with safe fallback

**What changed**
**Guards updated (6)**

deadLineGuard
allResponseTasksCompletedGuard
pcqGuard
checkYourAnswersClaimGuard
pcqGuardClaim
claimIssueTaskListGuard

**Controllers updated (6)**

checkAnswersController (response, GET)
requestMoreTimeController
responseDeadlineOptionsController
understandingYourOptionsController
agreedResponseDeadlineController
claim/checkAnswersController (GET)

**New files**

src/main/common/utils/claimRequestLocals.ts
Unit tests for helpers, guards, and controllers

```
[ ] Yes
[x] No
```


anirudha-deshpande (approved):

src/main/routes/features/response/responseDeadline/agreedResponseDeadlineController.ts
```diff
@@ -8,7 +8,8 @@ import {
 import {GenericForm} from 'form/models/genericForm';
 import {constructResponseUrlWithIdParams} from 'common/utils/urlFormatter';
 import {ResponseDeadlineService} from 'services/features/response/responseDeadlineService';
-import {generateRedisKey, getCaseDataFromStore} from 'modules/draft-store/draftStoreService';
+import {generateRedisKey} from 'modules/draft-store/draftStoreService';
```
pats-john, line 11: We do not seem to have used this import in this file.Is this necessary?

src/main/routes/features/response/responseDeadline/responseDeadlineOptionsController.ts
```diff
@@ -5,7 +5,8 @@ import {
   REQUEST_MORE_TIME_URL,
   RESPONSE_DEADLINE_OPTIONS_URL, APPLICATION_TYPE_URL,
 } from 'routes/urls';
-import {generateRedisKey, getCaseDataFromStore} from 'modules/draft-store/draftStoreService';
+import {generateRedisKey} from 'modules/draft-store/draftStoreService';
```
pats-john, line 8: We do not seem to have used this import in this file.Is this necessary?

src/main/routes/features/response/requestMoreTimeController.ts
```diff
@@ -1,6 +1,7 @@
 import {NextFunction, Request, RequestHandler, Response, Router} from 'express';
 import {RESPONSE_TASK_LIST_URL, REQUEST_MORE_TIME_URL, APPLICATION_TYPE_URL} from '../../urls';
-import {generateRedisKey, getCaseDataFromStore} from 'modules/draft-store/draftStoreService';
+import {generateRedisKey} from 'modules/draft-store/draftStoreService';
```
pats-john, line 3: We do not seem to have used this import in this file.Is this necessary?

pats-john (dismissed):

src/main/routes/features/response/requestMoreTimeController.ts
pats-john: This was the same import, but the getCaseDataFromStore removal shows it as newly added code.
Please ignore the comment.

PR #7937, opened 2026-07-16
DTSCCI-3841: refactor claimant response service

elvianihuseyin2:
### JIRA link ###

https://tools.hmcts.net/jira/browse/DTSCCI-3841

### Change description ###

new module: claimantResponseTaskListReset.ts

CLAIMANT_RESPONSE_TASK_LIST_RESET_RULES — single readable rule table
Named helpers for each side-effect (clearMediationCarm, clearTaskListData, clearAcceptOrRejectAmountData, etc.)
applyTaskListResetRules() — small reducer that evaluates rules in order (continue for independent rules; first exclusive match wins)
Service: saveClaimantResponse now calls applyTaskListResetRules instead of nested if/else mutation.

Tests: New claimantResponseTaskListReset.test.ts covers every rule branch, helper behaviour, and continue/stop semantics. Also fixed the full-defence “paid agreed” service test so it uses the real property path (hasFullDefenceStatesPaidClaimSettled) instead of relying on shared mutable state.

```
[ ] Yes
[x] No
```


src/main/services/features/claimantResponse/claimantResponseTaskListReset.ts
```diff
@@ -0,0 +1,168 @@
...
+  id: string;
+  when: (context: TaskListResetContext) => boolean;
+  apply: TaskListResetEffect;
+  /** When true, matching this rule does not stop evaluation of later rules. */
+  continue?: boolean;
+};
+
+const SUGGESTED_PAYMENT_INTENTION_CHILD_PROPERTIES = ['paymentOption', 'paymentDate', 'repaymentPlan'] as const;
+
+export function deleteClaimantResponseFields(claim: Claim, fields: readonly ClaimantResponseField[]): void {
```
gmikerlemos: Do these helpers and rule types need to be exported? They only seem to be used by this file and its unit tests. Keeping them private and testing the behaviour through applyTaskListResetRules would make the public API smaller. This is not blocking, just a suggestion.

gmikerlemos (approved):
Lgtm just a small  comment which is a non-blocking suggestion.

src/main/services/features/claimantResponse/claimantResponseTaskListReset.ts
elvianihuseyin2: updated

PR #7961, opened 2026-07-21
OCCC-285

sbagga-moj:

### JIRA link ###
https://tools.hmcts.net/jira/browse/OCCC-285

### Change description ###

Breathing Space (CUI) - Successful Entry

```
[ ] Yes
[x] No
```


src/main/common/models/breathingSpace/breathingSpaceEnterInfo.ts
```diff
@@ -0,0 +1,20 @@
+import {BreathingSpaceType} from 'models/breathingSpace/breathingSpaceType';
+
+export class BreathingSpaceEnterInfo {
+  type?: BreathingSpaceType;
+  reference?: string;
+  start?: Date;
+  expectedEnd?: Date | null;
```
toyin-hmcts, line 7: do we still need the expected end date since this is not part of the start flow anymore?

src/main/common/models/breathingSpace/breathingSpaceLiftInfo.ts
```diff
@@ -0,0 +1,7 @@
+export class BreathingSpaceLiftInfo {
```
toyin-hmcts, line 1: the exit flow also includes the optional "why was it lifted?" question - that should be a field here no?

toyin-hmcts (changes requested):

src/main/common/models/breathingSpace/breathingSpaceEnterInfo.ts
sbagga-moj: Yes, I have added it intentionally... once I merge it into the mfeature branch, then the same model can be used by Bharat/Gajanan. 

First, I deleted and later I realised that eventually we need this field, and I have added it back :-)

src/main/common/models/breathingSpace/breathingSpaceLiftInfo.ts
sbagga-moj: Same logic... when we translate from CCD to CUI, then we need both models. I have added this because we need this in the exit flow.

image

Added the screenshot for reference

src/main/common/models/breathingSpace/breathingSpaceEnterInfo.ts
toyin-hmcts: I think the end date question is only asked as part of the exit flow so this only needs to be part of BreathingSpaceLiftInfo (exit flow) and not the entry.

src/main/common/models/breathingSpace/breathingSpaceLiftInfo.ts
toyin-hmcts: screenshot is of the BreathingSpaceEnterInfo rather than BreathingSpaceLiftInfo. for the BreathingSpaceLiftInfo, there need to be a field to hold the why was it lifted answer.

toyin-hmcts (changes requested):

src/main/common/models/breathingSpace/breathingSpaceEnterInfo.ts
sbagga-moj: I spent some time and I forgot to mention that End Date (at least null) is needed to add a condition to show the lift link (including dashboard notification) on the dashboard. Without lift, I cannot add this condition

hasBreathingSpace(): boolean {
    return this.enterBreathing && !this.liftBreathing;
  }

this condition needed for AC3 in the story.

toyin-hmcts (approved):

PR #7972, opened 2026-07-23
DTSCCI-5964 Run reduced-stack create-claim test in Jenkins preview

gmikerlemos:
## Change

- add an opt-in reduced-stack Jenkins preview selected with the `pr-values:reducedStack` label
- deploy the real Civil Citizen UI and WireMock together in the PR preview namespace
- route the selected create-claim journey to preview WireMock while disabling Civil Service and its CCD, Camunda and supporting preview dependencies
- use CUI's existing `e2eTest` session and in-memory draft-store configuration; this change does not build or extend an IDAM stub
- package the consumer-owned Civil Service mappings in the CUI Helm chart and overlay them onto the shared mappings for local diagnostics
- run the WireMock-backed create-claim journey through the existing Jenkins functional-test stage
- verify the claim submission and submitted-claim lookup through WireMock's request journal
- fail the test when expected calls are missing or WireMock receives unmatched requests, and archive the request diagnostics with the functional-test artifacts
- install Playwright Chromium on the Jenkins VM that runs the functional stage, without relying on an agent cache
- leave the standard full-stack preview path unchanged when the reduced-stack label is absent
- retain `yarn test:mocked-functional` as a local developer diagnostic, not the authoritative CI execution

## Verification

- final implementation commit: `e0f570bb6038`
- Jenkins builds 25–36: 12 consecutive successful executions
- pass rate: 100% (12/12)
- median Jenkins wall-clock duration: 10m41s
- minimum/maximum duration: 9m08s / 17m55s
- every successful pipeline enforced the reduced-stack health gate, browser journey and WireMock request assertions
- expected WireMock claim-submission and submitted-claim lookup requests: received
- unmatched WireMock requests: zero
- `yarn test:mocked-functional`: passed locally, one create-claim scenario in approximately two minutes
- lint: passed with existing repository warnings and no errors
- JSON, YAML, shell syntax and diff validation: passed

Earlier runs are not included in the final reliability sample because they identified defects that were subsequently fixed. These included missing WireMock health/user-role mappings, an incompatible IDAM smoke test and Playwright being installed on a different Jenkins VM from the functional stage.

## Comparison with the full-stack path

The historical spike sample found 116 relevant Jenkins failures, of which 86 (74.1%) were conclusively caused by preview dependency setup before browser tests ran. The final reduced-stack sample completed 12/12 runs without Civil Service, CCD, Camunda, WA or their supporting preview services. A like-for-like successful full-stack duration baseline was not captured, so no unsupported numeric speed improvement is claimed.

## Usage

Apply `pr-values:reducedStack` to run the reduced-stack preview. Remove the label to restore the standard preview deployment. Do not combine it with `pr-values:fullDeployment`.

[Jira DTSCCI-5964](https://tools.hmcts.net/jira/browse/DTSCCI-5964)


sabahirfan (approved):
LGTM

anirudha-deshpande (approved):

PR #7973, opened 2026-07-23
DTSCCI-5961: set Node keepAliveTimeout above Traefik idle timeout to fix intermittent 502s

madhan-mahadevan:
Node's default keepAliveTimeout (5s) is below Traefik's 90s backend idleConnTimeout, so Node closes pooled sockets Traefik still reuses toward the pod, returning intermittent 502s. Set keepAliveTimeout=185000 on the http.Server so Traefik closes idle connections first. headersTimeout left at the Node 24 default (60s) - it governs header receipt, not idle keep-alive.

### DIFF
--- a/src/main/server.ts
+++ b/src/main/server.ts
@@ -28,6 +28,13 @@ process.on('unhandledRejection', (reason: any, promise: Promise<any>) => {
   
 });
 
+// Node's default keepAliveTimeout (5s) is below Traefik's 90s backend idleConnTimeout, so Node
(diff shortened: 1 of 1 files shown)


rajakm (approved):

gmikerlemos (approved):
Reviewed the change and the evidence on DTSCCI-5961. The timeout configuration looks correct, and the preview verification confirms the effective value is above Traefik's idle timeouts. No issues from me.

PR #8020, opened 2026-08-05
DTSCCI-5975 Add critical create-claim Pact coverage

gmikerlemos:
## Summary
- add consumer contracts for claim-fee lookup and create-claim submission
- assert only request fields, headers and response fields CUI consumes
- align the create-claim WireMock fee type with the verified Civil Service response

## Provider
- https://github.com/hmcts/civil-service/pull/8200

## Verification
- `yarn test:pact --runInBand` (4 suites, 8 tests passed)
- targeted ESLint passed
- Civil Service provider verification passed against the generated pact
- controlled number-vs-string incompatibility failed provider verification as expected

[Jira DTSCCI-5975](https://tools.hmcts.net/jira/browse/DTSCCI-5975)


src/main/common/models/civilClaimResponse.ts
```diff
@@ -336,7 +336,7 @@ export interface CCDClaim extends ClaimUpdate {
 }
 
 export interface ClaimFeeData {
-  calculatedAmountInPence?: number;
+  calculatedAmountInPence?: number | string;
```
elvianihuseyin2, line 339: do we still need to keep the number type? as it seems to be changed to string

elvianihuseyin2 (approved):

src/main/common/models/civilClaimResponse.ts
gmikerlemos: Good point. The number type is still required by the internal ClaimFeeData consumers, but widening the shared model to number | string made comparisons unsafe. I have separated the Civil Service response DTO (string) from the internal fee model (number) and now normalise the value at the CivilServiceClient boundary in commit 193181a72b. The Pact suite (4 suites / 8 tests), focused unit tests (3 suites / 93 tests), and targeted ESLint all pass.

ruban72 (approved):

